PR #3, opened 2019-10-28
Provide a specialized interval tree for resource management

jiangliu:
This PR implements a specialized version of interval tree for resource management, which could be used manage normal VMM resources.

### DIFF
--- a/src/interval_tree.rs
+++ b/src/interval_tree.rs
@@ -1,10 +1,47 @@
 // Copyright (C) 2019 Alibaba Cloud. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! A special interval tree implementation for VMM resource management.
+//! An interval tree implementation specialized for VMM resource management.
 //!
 //! It's not designed as a generic interval tree, but specialized for VMM resource management.
 //! In addition to the normal get/insert/delete/update operations, it also implements allocate/free.
+//!
+//! # Examples
+//! ```rust
+//! extern crate vm_allocator;
+//! use vm_allocator::{IntervalTree, Range, Constraint, NodeState};
+//!
+//! // Create an interval tree and add available resources.
+//! let mut tree = IntervalTree::<u64>::new();
+//! tree.insert(Range::new(0x100u32, 0x100u32), None);
+//! tree.insert(Range::new(0x200u16, 0x2ffu16), None);
+//!
+//! // Allocate a range with constraints.
+//! let mut constraint = Constraint::new(8u64);
+//! constraint.min = 0x211;
+//! constraint.max = 0x21f;
+//! constraint.align = 0x8;
+//!
+//! let key = tree.allocate(&constraint);
+//! assert_eq!(key, Some(Range::new(0x218u64, 0x21fu64)));
+//! let val = tree.get(&Range::new(0x218u64, 0x21fu64));
+//! assert_eq!(val, Some(NodeState::Allocated));
+//!
+//! // Associate data with the allocated range and mark the range as occupied.
+//! // Note: caller needs to protect from concurrent access between allocate() and the first call
+//! // to update() to mark range as occupied.
+//! let old = tree.update(&Range::new(0x218u32, 0x21fu32), 2);
+//! assert_eq!(old, None);
+//! let old = tree.update(&Range::new(0x218u32, 0x21fu32), 3);
+//! assert_eq!(old, Some(2));
+//! let val = tree.get(&Range::new(0x218u32, 0x21fu32));
+//! assert_eq!(val, Some(NodeState::Valued(&3)));
+//!
+//! // Free allocated resource.
+//! let old = tree.free(key.as_ref().unwrap());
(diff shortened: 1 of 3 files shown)


src/interval_tree.rs
```diff
@@ -349,6 +442,32 @@ impl<T> IntervalTree<T> {
...
+    /// ```rust
+    /// extern crate vm_allocator;
+    /// use vm_allocator::{IntervalTree, Range, Constraint};
+    ///
+    /// let mut tree = IntervalTree::<u64>::new();
+    /// tree.insert(Range::new(0x100, 0x100), None);
+    /// tree.insert(Range::new(0x200, 0x2ff), None);
+    ///
+    /// let constraint = Constraint::new(2);
+    /// let key = tree.allocate(&constraint);
```
liujing2: nit: Seems no Constraint definition in the first and second commit so far :)

src/interval_tree.rs
```diff
@@ -0,0 +1,449 @@
...
+    }
+}
+
+/// Internal tree node to implement interval tree.
+#[derive(Debug)]
+struct InnerNode<T> {
+    /// Interval handled by this node.
+    key: Range,
+    /// Optional contained data, None if the node is free.
+    data: Option<T>,
```
liujing2: What is `data` designed for? I can assume the usage, for example, IO address space [start, end] is described by `key` as [start, end], ioapic irq number N is described by `key` as [N, N].

jiangliu: It could be used to hold memfd for guest memory.

src/lib.rs
```diff
@@ -1,6 +1,43 @@
...
+    pub max: u64,
+    /// Alignment for the allocated resource.
+    pub align: u64,
+    /// Resource allocation policy.
+    pub policy: AllocPolicy,
+}
+
+impl Constraint {
+    /// Create a new constraint object with default settings.
+    pub fn new(size: u64) -> Self {
```
liujing2: It seems this function needs more parameters?

src/interval_tree.rs
liujing2: Got it. Thanks for explanation.

src/lib.rs
jiangliu: It's used to cover the most common usage scenario. For rare use cases, you may dirrectly construct a Constraint object or we may add some new interfaces for those cases.

liujing2: So what is the difference between this and `ResourceConstraint` defined in https://github.com/rust-vmm/vm-device/pull/9 ?

jiangliu: It's two related abstractions. With Device::ResourceConstraint and vm_allocator::Constraint, there's still one piece missing of the whole puzzle. We need the vmm to implement a resource manager, which converts Device::ResourceConstraint to vm_allocator::Constraint.
1) vm_allocator is just a generic resource allocator,
2) vmm::ResourceManager needs to define type of resources and use vm_allocator to manage its resources.
It's still unclear whether vmm::ResourceManager should be VMM specific or we need another vmm_resource crate.

liujing2: > It's two related abstractions. With Device::ResourceConstraint and vm_allocator::Constraint, there's still one piece missing of the whole puzzle. We need the vmm to implement a resource manager, which converts Device::ResourceConstraint to vm_allocator::Constraint.
> 
Ok, this is the "glue" part between vm-device and vm-allocator.

> 1. vm_allocator is just a generic resource allocator,
> 2. vmm::ResourceManager needs to define type of resources and use vm_allocator to manage its resources.
>    It's still unclear whether vmm::ResourceManager should be VMM specific or we need another vmm_resource crate.

I think we also need a conversion from the vmm::ResourceManager::allocate_resources() return type to vm_device::Resource for registering devices.

src/interval_tree.rs
liujing2: For other resources that don't need this `data`, what is the specific type of T?

jiangliu: A simple unit struct "()" should be enough.

src/lib.rs
jiangliu: Yes, that's true. Interval tree just manages a set of IDs, the resource manager needs to define the specific resources the ID stands for.

sameo:
Before looking deeper into the actual code, I guess my first general question would be: why couldn't we use one of the existing rust collection structure as the inner `IntervalTree` tree structure:

```rust
struct IntervalTree<T> {
          root: BTreeMap<Range, T>>,          
}
```

jiangliu:
> Before looking deeper into the actual code, I guess my first general question would be: why couldn't we use one of the existing rust collection structure as the inner `IntervalTree` tree structure:
> 
> ```rust
> struct IntervalTree<T> {
>           root: BTreeMap<Range, T>>,          
> }
> ```

IntervalTree supports using `range(start, end)` as the key, so we could search the entry containing a specific value(point).

AlexandruCihodaru:
Closing this as it is obsolete.

PR #9, opened 2022-01-11
IdAllocator documentation and implementation

AlexandruCihodaru:
- Added IdAllocator design and documentation
- Added IdAllocator initial implementation and unit tests

The documentation and implementation of AddressAllocator with IntervalTree backend will be added in a subsequent PR.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,4 @@
-# Add the list of code owners here (using their GitHub username)
+@AlexandruCihodaru
+@alsrdn
+@jiangliu
+@andreeaflorescu
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,10 @@
 [package]
-name = "crate-template"
+name = "vm-allocator"
 version = "0.1.0"
-authors = [TODO]
+authors = ["rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+description = "Resource allocators."
 license = "Apache-2.0 OR BSD-3-Clause"
 edition = "2018"
 
 [dependencies]
+libc = ">=0.2.39"
--- a/README.md
+++ b/README.md
@@ -1,41 +1,91 @@
-# Crate Name
+# vm-allocator
 
-## Design
+`vm-allocator` is a crate designed to to provide allocation and release strategies
+that are needed by the VMM during the lifetime of a virtual machine. Possible
+resource types that a VMM could allocate using vm-allocator are MMIO addresses,
+PIO addresses, GSI numbers, device ids.
 
-TODO: This section should have a high-level design of the crate.
+We have decided to have two allocator implementations, one for resources that can
+be abstracted to an integer and another allocator for addresses. We chose
+to have a separate allocator for addresses to add more semantic meaning to this
+resources (i.e. it needs informations like alignment which for resources like
+interrupt numbers are not needed). The main components are:
 
-Some questions that might help in writing this section:
-- What is the purpose of this crate?
-- What are the main components of the crate? How do they interact which each
-  other?
+- `IDAllocator` - which should be used for all resources that can be reduced to
+an integer type.
+- `AddressAllocator` - which should be used to allocate address ranges in different
+address spaces. This component is a wrapper over `IntervalTree` that adds semantics
+to address ranges.
 
-## Usage
+## ID Allocator
 
-TODO: This section describes how the crate is used.
+### Design
 
-Some questions that might help in writing this section:
-- What traits do users need to implement?
-- Does the crate have any default/optional features? What is each feature
-  doing?
-- Is this crate used by other rust-vmm components? If yes, how?
+This allocator should be used to allocate resources that can be reduced to an integer
+type like legacy GSI numbers or KVM memory slot IDs. The
+characteristics of such a resource are represented by the `IdAllocator` struct.
 
-## Examples
+The struct that defines the IdAllocator contains the ends of the interval that is
+managed, a field that points at the next available ID and a BTreeSet that is used
+to store the released IDs. We choosed to use a BTreeSet because the average
+complexity for deletion and insertion is O(logN) compared to Vec for example,
+another benefit is that the entries are sorted so we will always use the first
+available ID.
 
-TODO: Usage examples.
+#### Allocation policy
+
+When allocating a new ID we always try to return the smallest one available. To
+do that we first search in the BTreeSet for any ID that was released and if we
+cannot find anything there we return the next ID from the range that was never
+allocated.
 
 ```rust
-use my_crate;
-
-...
+/// Id allocator representation.
+pub struct IdAllocator {
+    // Begining of the range of IDs that we want to manage.
+    range_base: u32,
+    // First available id that was never allocated. 
+    next_id: Option<u32>,
+    // End of the range of IDs that we want to manage.
+    range_end: u32,
+    // Set of all freed ids that can be reused at subsequent allocations.
+    freed_ids: BTreeSet<u32>,
+}
 ```
+
+The `IdAllocator` struct implements methods for allocating and releasing IDs.
+
+```rust
+impl IdAllocator {
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_max`
+    pub fn new(range_base: u32, range_end: u32) -> std::result::Result<Self, Error> { }
+
+    /// Allocate an ID from the managed ranged.
+    pub fn allocate_id(&mut self) -> Result { }
+
+    /// Frees an id from the managed range.
+    pub fn free_id(&mut self, id: u32) -> Result { }
+}
+```
+
+### Usage
+
+Add vm-allocator as a dependency in Cargo.toml
+
+```toml
+[dependencies]
+vm-allocator = "*"
+````
+
+Then add extern crate vm-allocator; to projects crate root.
+The VMM using this crate should instantiate an IdAllocator object for each resource
+type they want to manage.
 
 ## License
 
-**!!!NOTICE**: The BSD-3-Clause license is not included in this template.
-The license needs to be manually added because the text of the license file
-also includes the copyright. The copyright can be different for different
-crates. If the crate contains code from CrosVM, the crate must add the
-CrosVM copyright which can be found
-[here](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/master/LICENSE).
-For crates developed from scratch, the copyright is different and depends on
-the contributors.
+This project is licensed under either of
+
+- [Apache License](http://www.apache.org/licenses/LICENSE-2.0), Version 2.0
+- [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
--- a/src/id_allocator.rs
+++ b/src/id_allocator.rs
@@ -1,0 +1,180 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright © 2019 Intel Corporation. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Provides allocation and releasing strategy for IDs.
+//!
+//! This module implements an allocation strategies for all resource types
+//! that can be abstracted to an integer.
+
+use crate::Error;
+use crate::Result;
+use std::collections::BTreeSet;
+
+/// Internal representation of IdAllocator. Contains the ends of the interval
+/// that is managed, a field that points at the next available ID, and a
+/// BTreeSet used to store the released IDs. The reason we chose a
+/// BTreeSet is that the average complexity for deletion and insertion is
+/// O(logN) compared to Vec for example, another benefit is that the entries
+/// are sorted so we will always use the first available ID.
+#[derive(Debug)]
+pub struct IdAllocator {
+    // Beginnbing of the range of IDs that we want to manage.
+    range_base: u32,
+    // First available id that was never allocated.
+    next_id: Option<u32>,
+    // End of the range of IDs that we want to manage.
+    range_end: u32,
+    // Set of all freed ids that can be reused at subsequent allocations.
+    freed_ids: BTreeSet<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_end`
+    pub fn new(range_base: u32, range_end: u32) -> std::result::Result<Self, Error> {
+        if range_end < range_base {
+            return Err(Error::InvalidRange(range_base, range_end));
+        }
+        Ok(IdAllocator {
+            range_base,
+            next_id: Some(range_base),
+            range_end,
+            freed_ids: BTreeSet::<u32>::new(),
+        })
+    }
+
+    fn id_in_range(&self, id: u32) -> bool {
+        // Check for out of range.
+        self.range_base <= id && id <= self.range_end
+    }
+
+    /// Allocate an ID from the managed ranged.
+    /// We first try to reuse one of the IDs released before. If there is no
+    /// ID to reuse we return the next available one from the managed range.
+    pub fn allocate_id(&mut self) -> Result {
+        // If the set containing all freed ids is not empty we extract the
+        // first entry from that set and return it.
+        if !self.freed_ids.is_empty() {
+            let ret_value = *self.freed_ids.iter().next().unwrap();
+            self.freed_ids.remove(&ret_value);
+            return Ok(ret_value);
+        }
+        // If no id was freed before we return the next available id.
+        if let Some(next_id) = self.next_id {
+            if next_id > self.range_end {
+                return Err(Error::ResourceExhausted);
+            }
+            // Prepare the next available id. If the addition overflows we
+            // set the next_id field to None and return Overflow at the next
+            // allocation request.
+            self.next_id = next_id.checked_add(1);
+            return Ok(next_id);
+        }
+        Err(Error::Overflow)
+    }
+
+    /// Frees an id from the managed range.
+    pub fn free_id(&mut self, id: u32) -> Result {
+        // Check if the id belongs to the managed range and if it was not
+        // released before. Return error if any of the conditions is not met.
+        if !self.id_in_range(id) {
+            return Err(Error::OutOfRange(id));
+        }
+        if let Some(next_id) = self.next_id {
+            if next_id < id {
+                return Err(Error::NeverAllocated(id));
+            }
+        }
+
+        // Insert the released id in the set of released id to avoid releasing
+        // it in next iterations.
+        self.freed_ids
+            .insert(id)
+            .then(|| id)
+            .ok_or(Error::AlreadyReleased(id))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::id_allocator::IdAllocator;
+    use crate::Error;
+
+    #[test]
+    fn test_slot_id_allocation() {
+        let faulty_allocator = IdAllocator::new(23, 5);
+        assert_eq!(faulty_allocator.unwrap_err(), Error::InvalidRange(23, 5));
+        let mut legacy_irq_allocator = IdAllocator::new(5, 23).unwrap();
+        assert_eq!(legacy_irq_allocator.range_base, 5);
+        assert_eq!(legacy_irq_allocator.range_end, 23);
+
+        let id = legacy_irq_allocator.allocate_id().unwrap();
+        assert_eq!(id, 5);
+        assert_eq!(legacy_irq_allocator.next_id.unwrap(), 6);
+
+        for _ in 1..19 {
+            assert!(legacy_irq_allocator.allocate_id().is_ok());
+        }
+
+        assert_eq!(
+            legacy_irq_allocator.allocate_id().unwrap_err(),
+            Error::ResourceExhausted
+        );
+    }
+
+    #[test]
+    fn test_u32_overflow() {
+        let mut allocator = IdAllocator::new(u32::MAX - 1, u32::MAX).unwrap();
+        assert_eq!(allocator.allocate_id().unwrap(), u32::MAX - 1);
+        assert_eq!(allocator.allocate_id().unwrap(), u32::MAX);
+        let res = allocator.allocate_id();
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), Error::Overflow);
+    }
+
+    #[test]
+    fn test_slot_id_free() {
+        let mut legacy_irq_allocator = IdAllocator::new(5, 23).unwrap();
+        assert_eq!(
+            legacy_irq_allocator.free_id(3).unwrap_err(),
+            Error::OutOfRange(3)
+        );
+        assert_eq!(legacy_irq_allocator.freed_ids.len(), 0);
+
+        for _ in 1..10 {
+            let _id = legacy_irq_allocator.allocate_id().unwrap();
+        }
+
+        let irq = 10;
+        legacy_irq_allocator.free_id(irq).unwrap();
+        assert!(legacy_irq_allocator.freed_ids.contains(&irq));
+        assert_eq!(
+            legacy_irq_allocator.free_id(10).unwrap_err(),
+            Error::AlreadyReleased(10)
+        );
+        let irq = 9;
+        legacy_irq_allocator.free_id(irq).unwrap();
+        assert_eq!(legacy_irq_allocator.freed_ids.len(), 2);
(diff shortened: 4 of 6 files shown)


src/id_allocator.rs
```diff
@@ -0,0 +1,201 @@
...
+
+pub type Result = result::Result<u32, Error>;
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::Error::*;
+
+        match *self {
+            Overflow => write!(f, "There are no more available ids in the specified range."),
+            InvalidInput => write!(f, "Specified id is not in the range."),
```
studychao, line 37: Should we print out the problematic id here to be more explicit?

src/id_allocator.rs
```diff
@@ -0,0 +1,201 @@
...
+            return Err(Error::InvalidInput);
+        }
+        Ok(id)
+    }
+
+    /// Allocate an ID from the managed ranged.
+    pub fn allocate_id(&mut self) -> Result {
+        // If the set containing all freed ids is not empty we extract the
+        // first entry from that set and return it.
+        if self.freed_ids.len() != 0 {
```
studychao, line 98: Could this change to "if !self.freed_ids.is_empty()" to be more straightforward?

src/id_allocator.rs
```diff
@@ -0,0 +1,201 @@
...
+            return Ok(ret_value);
+        }
+        // If no id was freed before we return the next available id.
+        let id = self.next_id;
+        if id > self.range_end {
+            return Err(Error::Overflow);
+        }
+        // Prepare the next available id and return `Error::Overflow` if the
+        // addition results in an overflow.
+        self.next_id = self.next_id.checked_add(1).ok_or(Error::Overflow)?;
```
studychao, line 110: Above two Overflow Error may be confusing to the user since it's hard to know whether the next_id is overflow or exceed the range_end.
Perhaps add another explicit error type for exceeding the range_end?

src/id_allocator.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Frees an id from the managed range.
+    pub fn free_id(&mut self, id: u32) -> Result {
+        // Check if the id belongs to the managed range and if it was not
+        // released before. Return error if any of the conditions is not met.
+        if self.check_id(id).is_err() || self.next_id < id {
+            return Err(Error::InvalidInput);
+        }
+        // Insert the released id in the set of released id to avoid releasing
+        // it in next iterations.
+        let _ = self.freed_ids.insert(id);
```
studychao, line 123: Could this change to "self.freed_ids.insert(id);" to be more straightforward?

src/id_allocator.rs
```diff
@@ -0,0 +1,201 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Provides allocation and releasing strategy for IDs.
+//! 
+//! This module implements an allocation strategies for all resource types
+//! that can be abstracted to an integer.
+
+
+//! Provides emulation for a minimal ARM PL031 Real Time Clock.
```
studychao, line 10: Is this part needed?:)

studychao (commented):
Hi, I have added some code review. Please take a look :)

src/id_allocator.rs
AlexandruCihodaru: Nope it was a miss from my part. Thank you for pointing this out.

src/id_allocator.rs
AlexandruCihodaru: Sure, good suggestion

CODEOWNERS
```diff
@@ -1 +1,4 @@
-# Add the list of code owners here (using their GitHub username)
+AlexandruCihodaru
```
studychao, line 1: should this part follow github's standard? You may need an @ before each github username
(https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)

README.md
```diff
@@ -1,41 +1,80 @@
...
+PIO addresses, IRQ numbers, MSI vectors, device ids.
 
-TODO: This section should have a high-level design of the crate.
-
-Some questions that might help in writing this section:
-- What is the purpose of this crate?
-- What are the main components of the crate? How do they interact which each
-  other?
+Main components are:
+- `IDAllocator` - which should be used for all resourcers that can be reduced to
```
studychao, line 9: typo?~

README.md
```diff
@@ -1,41 +1,80 @@
...
-- Does the crate have any default/optional features? What is each feature
-  doing?
-- Is this crate used by other rust-vmm components? If yes, how?
+```toml
+[dependencies]
+vm-allocator = "*"
+````
+Then add extern crate vm-allocator; to projects crate root.
+The VMM using this crate should instantiate an IdAllocator object for each resource
+type thei want to manage.
```
studychao, line 56: typo?

src/lib.rs
```diff
@@ -1,6 +1,8 @@
...
-//! Dummy crate needs high-level documentation.
-/// Dummy public function needs documentation.
-pub fn it_works() {
-    assert!(true);
-}
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//! Manages system resources that can be allocated to VMs and their devices.
+
+extern crate libc;
```
studychao, line 5: https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html

due to rust 2018 edition guide, there is no need for extern crate. Please check the document to see if it's needed in your circumstance:)

studychao (commented):
I've added some few more comments, please take a look at them:)

studychao (dismissed):

README.md
```diff
@@ -1,41 +1,80 @@
-# Crate Name
+# vm-allocator
 
-## Design
+`vm-allocator` is a crate designed to to provide allocation and release strategies
```
andreeaflorescu, line 3: I think one relevant thing that is missing from this document is the design choice of having one allocator per resource. We should somehow write about this as well.

src/id_allocator.rs
```diff
@@ -0,0 +1,199 @@
...
+    range_end: u32,
+    // Set of all freed ids that can be reused at subsequent allocations.
+    pub freed_ids: BTreeSet<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_max`
+    pub fn new(range_base: u32, range_max: u32) -> Self {
```
andreeaflorescu, line 73: I find u64 to be more future proof because then we could easily re-use the code for addresses as well. We can leave it as is for now though.

andreeaflorescu (commented):
Thanks for splitting this effort in smaller PRs that are easy to review! :confetti_ball: 

I did an initial round of review, and I have some question. I still have to look at the tests, and the allocate_id function can be optimized a bit I believe. That's nothing that can't be changed later though.

Cargo.toml
```diff
@@ -1,8 +1,11 @@
 [package]
-name = "crate-template"
+name = "vm-allocator"
 version = "0.1.0"
-authors = [TODO]
+authors = ["rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+description = "Allocation of resources needed by VMM"
+keywords = ["Id", "Address space", "IRQ", "MSI", "MSI-x"]
```
andreeaflorescu, line 6: You can't use space in keywords, it will fail when you'll try to publish it. I also think the keywords are not speaking so much about what the crate does.

I am thinking about a keyword that speaks more about the main action of the crate "allocating resources" than about what is the type of resources that you can allocate. At this point, I can't come up with anything.

src/id_allocator.rs
```diff
@@ -0,0 +1,199 @@
...
+            range_base,
+            next_id: range_base,
+            range_end: range_max,
+            freed_ids: BTreeSet::<u32>::new(),
+        }
+    }
+
+    /// Checks if a specified id belongs to the range that this instance of
+    /// `IdAllocator` is managing and if it was not freed before.
+    fn check_id(&self, id: u32) -> Result {
```
andreeaflorescu, line 84: nit: Can we rename this to `id_in_range` instead? That would make the comment unnecessary and it speaks more to what the function is doing.

src/id_allocator.rs
```diff
@@ -0,0 +1,199 @@
...
+use std::fmt;
+use std::result;
+
+// Error type for IdAllocator usage.
+#[derive(Debug)]
+pub enum Error {
+    // All ids from the range specified are exhausted.
+    Overflow,
+    // An id that is not part of the specified range was requested to be released.
+    InvalidInput(u32),
```
andreeaflorescu, line 19: If this is only used for out of range IDs, can we rename this to `OutOfRange` or similar so that the error variant is more verbose?

src/id_allocator.rs
andreeaflorescu: This logic is a bit problematic because it does not allow you to reserve the last ID in the range if the range happens to end at u32::MAX. Can we instead keep the `last_used` and increment it when a new allocation request comes in? Then the error would be return only for the operation that is problematic (in this case u32::MAX + 1).

README.md
```diff
@@ -1,41 +1,80 @@
...
+for address ranges.
+## ID Allocator
+### Design
+This allocator should be used to allocate resources that can be reduced to an integer
+type like legacy IRQ numbers, MSI interrupt numbers or KVM memory slot IDs. The
+characteristics of such a resource are represented by the `IdAllocator` struct.
+```rust
+/// Id allocator representation.
+pub struct IdAllocator {
+    // Begining of the range of IDs that we want to manage.
```
andreeaflorescu, line 22: nit: typo

src/id_allocator.rs
```diff
@@ -0,0 +1,199 @@
...
+/// }
+/// let irq = 8;
+/// let freed_id_result = legacy_irq_allocator.free_id(irq);
+/// if freed_id_result.is_err() {
+///     println!("Failed to free the specified id: {:?}", freed_id_result);
+/// }
+/// ```
+
+/// Id allocator representation.
+pub struct IdAllocator {
```
andreeaflorescu, line 58: Can you add a few comments about the internal design choices of this structure? What I am finding difficult to understand is why do we need to use an interval tree as oppose to just a vector in which we keep the freed ids. I would assume that some trade-offs were made and that these advantages and disadvantages were put in balance, but this kind of information tends to be forgotten, so it's better to keep it next to the code. What I don't particularly like about BTree is Rust is that it's making you implement useful operations that are not yet stabilized (the one that would be useful for us is `pop_first`).

Also, the allocation strategy should be mentioned somewhere IMHO (i.e. we first try to reserve IDs from previously freed ones), but I am not sure what is the best way to deliver the message.

src/id_allocator.rs
AlexandruCihodaru: It is not used only for out of range IDs. This is also returned when we try to release an ID that was never allocated or already released.

src/id_allocator.rs
```diff
@@ -0,0 +1,199 @@
...
+                write!(f, "There are no more available ids in the specified range.")
+            }
+        }
+    }
+}
+
+/// # Example
+///
+/// ```rust
+/// # use vm_allocator::IdAllocator;
```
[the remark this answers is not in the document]
AlexandruCihodaru: done! ptal.

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+    InvalidInput(u32),
+    /// There are no more IDs available in the manage range
+    ResourceExhausted,
+}
+
+pub type Result = result::Result<u32, Error>;
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::Error::*;
```
gsserge, line 28: nit: `self::` is superfluous here.

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+//! This module implements an allocation strategies for all resource types
+//! that can be abstracted to an integer.
+
+use std::collections::BTreeSet;
+use std::fmt;
+use std::result;
+
+/// Error type for IdAllocator usage.
+#[derive(Debug, PartialEq)]
+pub enum Error {
```
gsserge, line 15: We might want to add an impl for `std::error::Error`. This costs us nothing (the impl will be empty), but allows the client code use more flexible error handling with boxed errors if needed.

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub fn new(range_base: u32, range_max: u32) -> Self {
+        IdAllocator {
+            range_base,
+            next_id: range_base,
+            range_end: range_max,
+            freed_ids: BTreeSet::<u32>::new(),
+        }
+    }
+
+    fn id_in_range(&self, id: u32) -> Result {
```
gsserge, line 88: Can this function return `bool` instead of `Result`?

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+/// }
+/// let irq = 8;
+/// let freed_id_result = legacy_irq_allocator.free_id(irq);
+/// if freed_id_result.is_err() {
+///     println!("Failed to free the specified id: {:?}", freed_id_result);
+/// }
+/// ```
+
+/// Internal representation of IdAllocator. Contains the ends of the interval
+/// that is managed, a field that points at the next available ID, and a binary
```
gsserge, line 59: nit: BTree is not a binary tree.

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+/// tree set used to store the released IDs. The reason we have chosed a
+/// BTreeSet is that the overage complexity for deletion and insertion is
+/// O(logN) compared to Vec for example, anothe benefit is that the entries
```
gsserge, line 62: typo: anothe -> another.

README.md
```diff
@@ -1,41 +1,95 @@
...
-- Does the crate have any default/optional features? What is each feature
-  doing?
-- Is this crate used by other rust-vmm components? If yes, how?
+This allocator should be used to allocate resources that can be reduced to an integer
+type like legacy IRQ numbers, MSI interrupt numbers or KVM memory slot IDs. The
+characteristics of such a resource are represented by the `IdAllocator` struct.
+
+```rust
+/// Id allocator representation.
+pub struct IdAllocator {
```
gsserge, line 26: If we duplicate part of the implementation in README, then we need to make sure to keep them in sync if something changes (method signatures, doc comments, etc).

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+    range_end: u32,
+    // Set of all freed ids that can be reused at subsequent allocations.
+    freed_ids: BTreeSet<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_max`
+    pub fn new(range_base: u32, range_max: u32) -> Self {
```
gsserge, line 79: We should use consistent naming for the range end struct field and parameter; now it's `range_end` in the struct and `range_max` in the constructor.

src/id_allocator.rs
andreeaflorescu: Can we have another error for that one instead? `InvalidInput` is a bit too generic because most of errors are caused by an invalid input in the end.

README.md
andreeaflorescu: We can also just move the examples either in the code, or in an `examples` directory. A layout example is available here: https://doc.rust-lang.org/cargo/guide/project-layout.html

Cargo.toml
```diff
@@ -1,8 +1,11 @@
 [package]
-name = "crate-template"
+name = "vm-allocator"
 version = "0.1.0"
-authors = [TODO]
+authors = ["rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+description = "Allocation of resources needed by VMM"
```
jiangliu, line 5: Resource allocators?

README.md
```diff
@@ -1,41 +1,95 @@
...
+    // Begining of the range of IDs that we want to manage.
+    range_base: u32,
+    // First available id that was never allocated. 
+    pub next_id: u32,
+    // End of the range of IDs that we want to manage.
+    range_end: u32,
+    // Set of all freed ids that can be reused at subsequent allocations.
+    pub freed_ids: BTreeSet<u32>,
```
jiangliu, line 34: why make these two fields as pub?

README.md
```diff
@@ -1,41 +1,95 @@
...
+    msi_interrupt_number_allocator: IdAllocator,
+    kvm_mem_slot: IdAllocator,
+}
+```
+
+Then the VMM should use the defined structure anytime it wants to allocate a new resource
 
-...
+```rust
+pub fn function_that_needs_resources(allocator: VMMDefinedAllocatorStruct) {
```
jiangliu, line 84: &mut VMMDefinedAllocatorStruct?

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+/// # Example
+///
+/// ```rust
+/// # use vm_allocator::IdAllocator;
+///
+/// let mut legacy_irq_allocator = IdAllocator::new(5, 23);
+///
+/// let irq_nr = legacy_irq_allocator.allocate_id();
+/// if irq_nr.is_err() {
+///     println!("Interrupt number allocation failed: {:?}", irq_nr);
```
jiangliu, line 49: Will the test case reported as failure when `irq_nr.is_err()` is true?
Seems it will break "cargo test".

src/id_allocator.rs
```diff
@@ -0,0 +1,204 @@
...
+/// }
+/// let irq = 8;
+/// let freed_id_result = legacy_irq_allocator.free_id(irq);
```
jiangliu, line 52: What happens if an unallocated id is freed?

src/lib.rs
```diff
@@ -1,6 +1,9 @@
...
 #![deny(missing_docs)]
-//! Dummy crate needs high-level documentation.
-/// Dummy public function needs documentation.
-pub fn it_works() {
-    assert!(true);
-}
+
+mod id_allocator;
+
+pub use crate::id_allocator::{Error as IdAllocatorError, IdAllocator};
```
jiangliu, line 9: It may help to simplify the implementation to share a common Error definition at the crate level.

jiangliu:
It would be great to mention the origin of the design/algorithm, either from Cloud Hypervisor or crosvm:)

src/id_allocator.rs
AlexandruCihodaru: If an ID that was not allocated before we will return an error. In this [line](https://github.com/rust-vmm/vm-allocator/blob/bde96f123e5d56e285a393d5459e4c0a0f48b634/src/id_allocator.rs#L122) we check to see if the id requested is in the range we are working with. We explicitly check that the ID was allocated `self.next_id < id` if no we return `InvalidInput` for now but I think the error name will change.

src/lib.rs
```diff
@@ -1,6 +1,99 @@
...
+    /// An id  that was never allocated was requested to be released.
+    NeverAllocated(u32),
+    /// There are no more IDs available in the manage range
+    ResourceExhausted,
+    /// The range to manage is invalid.
+    InvalidRange(u32, u32),
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
gsserge, line 31: I might be wrong here, but I think that there's no need to define the `source()` method. The default implementation  that returns `None` is more appropriate, because the error itself should not be its own source. The source would be appropriate if, say, there was an I/O error leading to this one.

src/id_allocator.rs
```diff
@@ -0,0 +1,182 @@
...
+    // Set of all freed ids that can be reused at subsequent allocations.
+    freed_ids: BTreeSet<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_max`
+    pub fn new(range_base: u32, range_end: u32) -> std::result::Result<Self, Error> {
+        if range_end <= range_base {
```
gsserge, line 37: I think the check should be `range_end < range_base`. the `allocate_id()` code suggests that the range is inclusive, therefore `IdAllocator::new(1,1);` should be allowed, because it contains a range with one element.

src/id_allocator.rs
```diff
@@ -0,0 +1,182 @@
...
+    // End of the range of IDs that we want to manage.
+    range_end: u32,
+    // Set of all freed ids that can be reused at subsequent allocations.
+    freed_ids: BTreeSet<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_max`
```
gsserge, line 35: `range_end` instead of `range_max`.

src/lib.rs
```diff
@@ -1,6 +1,98 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::Error as AllocError;
+    use std::error::Error;
+
+    #[test]
+    fn test_error_messages() {
```
andreeaflorescu, line 68: We avoid these kind of coverage tests in rust-vmm because we find them to break easily and don't bring a lot of value otherwise. Would you mind removing them if you don't feel strongly about it?

src/id_allocator.rs
```diff
@@ -0,0 +1,182 @@
...
+        // Unwrap is safe because we already checked that next_id is not None.
+        let id = self.next_id.unwrap();
+        if id > self.range_end {
+            return Err(Error::ResourceExhausted);
+        }
+        // Prepare the next available id and return `Error::Overflow` if the
+        // addition results in an overflow.
+        // Unwrap is safe because we already checked that next_id is not None.
+        self.next_id = self.next_id.unwrap().checked_add(1);
+        Ok(id)
```
georgepisaltu, line 77: I'd do this with a `match` or an `if let` to avoid the `unwrap` below.

src/id_allocator.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+        // Unwrap is safe because we already checked that next_id is not None.
+        let id = self.next_id.unwrap();
+        if id > self.range_end {
+            return Err(Error::ResourceExhausted);
+        }
+        // Prepare the next available id and return `Error::Overflow` if the
+        // addition results in an overflow.
+        // Unwrap is safe because we already checked that next_id is not None.
+        self.next_id = self.next_id.unwrap().checked_add(1);
```
georgepisaltu, line 76: If we use a `match` or `if let` above, we can use a `map` here to avoid the `unwrap`.

AlexandruCihodaru:
@georgepisaltu  @andreeaflorescu  @gsserge  @studychao  @jiangliu  thank you for your reviews, I have addressed your comments, please take another look.

src/lib.rs
gsserge: There's no need to manually re-implement the default implementation of [source](https://doc.rust-lang.org/std/error/trait.Error.html#method.source). The empty impl is sufficient:
```rust
impl std::error::Error for Error {}
```

Cargo.toml
gsserge: This comment about spaces in [keywords](https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field) has not been addressed.

gsserge:
Two small comments regarding the keywords and the Error impl, otherwise looks good to me.

Cargo.toml
AlexandruCihodaru: I have removed keywords to not block this PR in this. We will add it when we find something to put there.

gsserge (approved):

georgepisaltu (approved):
LGTM!

gsserge (approved):

georgepisaltu (approved):

studychao (approved):

studychao:
LGTM:)

jiangliu (approved):

dwmw2:
> It would be great to mention the origin of the design/algorithm, either from Cloud Hypervisor or crosvm:)

"great"? 

I think that such a mention is *mandatory*, for submissions which are derived from other work.

Obviously, for basic licensing reasons, if we re-use any other work we need to *know* what we're using and where it comes from, and be sure that the licence is compatible. When an engineer adds their `Signed-off-by:` to a commit, that is what they are promising.

So it's already absolutely required that they *know* the origin and didn't just paste random stuff they found on Stack Overflow, or the whole project is in jeopardy. Given that, it costs nothing — and demonstrates that they have done their necessary due diligence — to expect them to state it. And is basic courtesy.

If this isn't already documented as a fundamental *requirement* for the project, it should be.

(Although it should also go without saying, as it's just basic engineer discipline.)

PR #12, opened 2022-02-02
Address allocator documentation

AlexandruCihodaru:
* Added AddressAllocator design and documentation.

The implementation of AddressAllocator with IntervalTree backend will be added in multiple subsequent PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,86 @@ impl IdAllocator {
 }
 ```
 
+## AddressAllocator
+
+### Design address allocator
+
+This allocator should be used to allocate memory slots, either for memmory-mapped
+I/O or for port I/O or any other kind of address space.
+
+The struct that defined the AddressAllocator contains the ends of the interval
+that is managed and an instance of `interval_tree` as an allocation engine. We
+chose to use `interval_tree` for memory slots allocation because it provides
+better creation performance and more query speed.
+
+```rust
+/// Address allocator representation
+pub struct AddressAllocator {
+    // Begining of the address space that we want to manage.
+    base: u64,
+    // End of the address space that we want to manage.
+    end: u64,
+    // Allocation engine, the allocation logic that is interfaced
+    // by AddressAllocator
+    interval_tree: IntervalTree,
+}
+````
+
+The `AddressAllocator` struct implements methods for allocating and releasing
+memory slots from the managed address space.
+
+```rust
+impl AddressAllocator {
+    ///  Creates a new AddressAllocator object with an empty IntervalTree
+    pub fn new(base: u64, size: u64) -> std::result::Result<Self, Error> { }
+
+    /// Allocate a resource range according the allocation constraints.
+    pub fn allocate(
+        &mut self,
+        _address: Option<u64>,
+        size: u64,
+        align_size: Option<u64>,
+        alloc_policy: AllocPolicy,
+    ) -> Result<Range> { }
+
+    /// Free an allocated range.
+    pub fn free(&mut self, key: &Range) -> Result<()> { }
+```
+
+Struct Constraint is used to describe the overall information of the resource needed to be allocated.
+
+```rust
+/// Struct to describe resource allocation constraints.
+pub struct Constraint {
+    /// Size to allocate.
+    pub size: u64,
+    /// Lower boundary for the allocated resource.
+    pub min: u64,
+    /// Upper boundary for the allocated resource.
+    pub max: u64,
+    /// Alignment for the allocated resource.
+    pub align: u64,
+    /// Resource allocation policy.
+    pub policy: AllocPolicy,
+}
+
+impl Constraint {
+    /// Create a new constraint object with default settings.
+    pub fn new(size: u64) -> Self { }
+
+    /// Set the min constraint.
+    pub fn min(mut self, min: u64) -> Self { }
+
+    /// Set the max constraint.
+    pub fn max(mut self, max: u64) -> Self { }
+
+    /// Set the alignment constraint.
+    pub fn align(mut self, align: u64) -> Self { }
+
+    /// Set the allocation policy.
+    pub fn policy(mut self, policy: AllocPolicy) -> Self { }
+```
+
 ### Usage
 
 Add vm-allocator as a dependency in Cargo.toml
--- a/src/allocation_engine/README.md
+++ b/src/allocation_engine/README.md
@@ -1,0 +1,135 @@
+# Interval Tree allocation engine
+
+## Design
+
+This interval tree implementation is defined to work as an allocation
+engine for memory-mapped I/O and port I/O address space. It provides
+underlaying methods for allocating, releaseing and updating a memory
+slot. This implementation is going to be used by address allocator
+component of vm-allocator, its main goal is to make address slots
+allocation easier to track and maintain.
+
+```rust
+pub struct IntervalTree {
+    pub(crate) root: Option<InnerNode>,
+}
+
+/// Internal tree node to implement interval tree.
+#[derive(Debug, PartialEq)]
+pub(crate) struct InnerNode {
+    /// Interval handled by this node.
+    pub(crate) key: Range,
+    /// Optional contained data, None if the node is free.
+    pub(crate) node_state: NodeState,
+    /// Optional left child of current node.
+    pub(crate) left: Option<Box<InnerNode>>,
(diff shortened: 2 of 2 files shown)


AlexandruCihodaru:
I have pushed a reference implementation [here](https://github.com/AlexandruCihodaru/vm-allocator/commit/5a8875eb6e84aae4be8ef81264ea20ca473083a7). Please take a look.

README.md
```diff
@@ -70,6 +70,86 @@ impl IdAllocator {
...
+## AddressAllocator
+
+### Design address allocator
+
+This allocator should be used to allocate memory slots, either for memmory-mapped
+I/O or for port I/O or any other kind of address space.
+
+The struct that defined the AddressAllocator contains the ends of the interval
+that is managed and an instance of `interval_tree` as an allocation engine. We
+chosed to use `interval_tree` for memory slots allocation because it provides
```
gsserge, line 82:
```suggestion
chose to use `interval_tree` for memory slots allocation because it provides
```

README.md
```diff
@@ -70,6 +70,86 @@ impl IdAllocator {
...
+}
+````
+
+The `AddressAllocator` struct implements methods for allocating and releasing
+memory slots from the managed address space.
+
+```rust
+impl AddressAllocator {
+    ///  Creates a new AddressAllocator object with an empty IntervalTree
+    pub fn new(base: u64, size: u64) -> std::result::Result<Self, Error> { }
```
gsserge, line 104: The struct has the `end` field, while the constructor has `size`. Is this by design, i.e. does `base` and `size` in the constructor get translated to `base` and `end` in the struct?

gsserge: Also, looking at the `allocate()` method below, the return type should probably be `Result<Self>`.

README.md
```diff
@@ -70,6 +70,86 @@ impl IdAllocator {
...
+better creation performance and more query speed.
+
+```rust
+/// Address allocator representation
+pub struct AddressAllocator {
+    // Begining of the address space that we want to manage.
+    base: u64,
+    // End of the address space that we want to manage.
+    end: u64,
+    // Allocation engine, the allocation logic that is interfaced
```
gsserge, line 92: "the ... logic that is interfaced" sounds a bit weird. Maybe just "The Interval Tree allocation engine"?

README.md
```diff
@@ -70,6 +70,86 @@ impl IdAllocator {
...
+    ///  Creates a new AddressAllocator object with an empty IntervalTree
+    pub fn new(base: u64, size: u64) -> std::result::Result<Self, Error> { }
+
+    /// Allocate a resource range according the allocation constraints.
+    pub fn allocate(
+        &mut self,
+        _address: Option<u64>,
+        size: u64,
+        align_size: Option<u64>,
+        alloc_policy: AllocPolicy,
```
gsserge, line 112: `policy: AllocPolicy` is probably sufficient, from the method's name it's pretty clear that it's about allocation.

README.md
```diff
@@ -70,6 +70,86 @@ impl IdAllocator {
...
+    /// Resource allocation policy.
+    pub policy: AllocPolicy,
+}
+
+impl Constraint {
+    /// Create a new constraint object with default settings.
+    pub fn new(size: u64) -> Self { }
+
+    /// Set the min constraint.
+    pub fn min(mut self, min: u64) -> Self { }
```
gsserge, line 141: I think that using just `self` as the receiver is sufficient (without `mut`).

README.md
AlexandruCihodaru: The allocate method return an interval [a, b] that correspond to the constraint, I do not believe it should return Result<Self>

gsserge: I see how my comment was confusing, sorry :( What I meant is that `new()` should return `Result<Self>` similar to how `allocate()` returns `Result<Range>`. So, `Result<Self>` instead of `std::result::Result<Self, Error>`.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+pub struct IntervalTree {
+    pub(crate) root: Option<InnerNode>,
+}
+
+/// Internal tree node to implement interval tree.
+#[derive(Debug, PartialEq)]
+pub(crate) struct InnerNode {
+    /// Interval handled by this node.
+    pub(crate) key: Range,
+    /// Optional contained data, None if the node is free.
```
gsserge, line 22: Probably a leftover from the original implementation when it was `Option`. Now it's `NodeState` enum, so no "Optional" and "None".

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+This interval tree implementation is defined to work as an allocation
+engine for memory-mapped I/O and port I/O address space. It provides
+underlaying methods for allocating, releaseing and updating a memory
+slot. This implementation is going to be used by address allocator
+component of vm-allocator, its main goal is to make address slots
+allocation easier to track and maintain.
+
+```rust
+pub struct IntervalTree {
+    pub(crate) root: Option<InnerNode>,
```
gsserge, line 14: Do we need to have the `pub(crate)` visibility here? Is it possible to completely hide away this implementation detail?

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+}
+
+pub enum NodeState {
+    /// Node is free
+    Free,
+    /// Node is allocated but without associated data
+    Allocated,
+}
+
+/// Insert the ket into the interval tree, returns Error if intersects with existing nodes.
```
gsserge, line 41: The methods should probably be wrapped into the `impl IntervalTree` block.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+pub fn insert(&mut self, key: Range, node_state: NodeState) -> Result<()> { }
+
+/// Update an existing entry and return the old value.
+pub(crate) fn update(&mut self, key: &Range, node_state: NodeState) -> Result<()> { }
```
gsserge, line 45: Why `pub(crate)` instead of `pub` like the rest of the methods?

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+pub fn delete(&mut self, key: &Range) -> Option<()> { }
+
+ /// Construct a new empty IntervalTree.
+pub fn new() -> Self { }
+
+/// Check whether the interval tree is empty.
+pub fn is_empty(&self) -> bool { }
+
+/// Get the data item associated with the key, or return None if no match found.
+pub fn get(&self, key: &Range) -> Option<NodeState> { }
```
gsserge, line 57: Just to clarify, `NodeState` is the actual data item, right?

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+    end: u64,
+    interval_tree: IntervalTree,
+}
+
+impl AddressAllocator {
+    pub fn new(base: u64, size: u64) -> std::result::Result<Self, Error> {
+        let interval_tree = IntervalTree::new();
+        let mut address_allocator = AddressAllocator {
+            base: base,
+            end: end.unwrap(),
```
gsserge, line 98: Where does `end` come from and why `.unwrap()`? Same question for the `allocate()` below.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,135 @@
...
+        &mut self,
+        _address: Option<u64>,
+        size: u64,
+        align_size: Option<u64>,
+        alloc_policy: AllocPolicy,
+    ) -> Result<Range> {
+        let constraint = Constraint::new(size).align(alignment).policy(alloc_policy);
+        // some other code here
+        if logical_condition {
+            let key = self.interval_tree.find_candidate(&constaint);
```
gsserge, line 120: The `find_candidate()` method was not in the list of the methods above.

gsserge:
Maybe it's just me, but I do have a strong feeling that we are not talking documentation here, but we are actually discussing the interfaces. It's also kinda awkward to review when the interfaces are part of the documentation. Maybe it would be better to have a very short docs with high-level concepts expressed with words, and the interfaces be part of the actual code.

AlexandruCihodaru:
As @gsserge  suggested [here](https://github.com/rust-vmm/vm-allocator/pull/12#issuecomment-1028162708) I have created a new pull request with the initial documentation and structs definitions. Will close #12 as #13 was opened.

PR #13, opened 2022-02-05
Interval tree initial documentation

AlexandruCihodaru:
* Add definition of initial structures that are needed by `IntervalTree`.
* Add initial documentation of `IntervalTree` allocation engine. 
 Created a new PR as suggested by @gsserge [here](https://github.com/rust-vmm/vm-allocator/pull/12#issuecomment-1028162708)


src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+pub struct Range {
+    pub min: u64,
+    pub max: u64,
+}
+
+// Node state for interval tree nodes.
+///
+/// Valid state transition:
+/// - None -> Free: IntervalTree::insert()
+/// - Free -> Allocated: IntervalTree::allocate()
```
gsserge, line 79: It seems like there's no separate `allocate()` method, but `insert(key, NodeState::Allocated)` instead.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+    /// Allocate first matched entry from the end of the range.
+    LastMatch,
+    // Allocate a memory slot starting with the specified addrres
+    // if it is available.
+    ExactMatch,
+}
+```
+
+Struct `Constraint` is used to describe the overall information of the resource
+ needed to be allocated. This structure is also used by IntervalTree to know where
```
gsserge, line 36: Unnecessary extra space on this and the next line.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+## Interval tree
+
+An interval tree is a tree data structure used for storing information about intervals.
+Specifically, it allows one to efficiently identify intervals that are overlaping
+with a given point, or another interval. We considered that this characteristic
+makes this data structure appropriate to be used as an allocation engine for
+memory slots inside an address space. The time complexity of an interval tree,
+namely O(log ⁡n+m) for  queries, O(log n) for creation and O(log n) for insertion
```
gsserge, line 62:
```suggestion
namely O(log ⁡n+m) for queries, O(log n) for creation and O(log n) for insertion
```

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+}
+```
+
+## Usage
+
+The concept of Interval Tree may seem complicated, but using vm-allocator to do
+resource allocation and release is simple and straightforward.
+
+To use the `IntervalTree` implementation as an address allocator one should first
+ create an interval tree  object and give an address space as a root node.
```
gsserge, line 112:
```suggestion
create an interval tree object and give an address space as a root node.
```

src/allocation_engine/mod.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2022 Alibaba Cloud. All rights reserved.
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2
+
+mod interval_tree;
+
+pub use interval_tree::{AllocPolicy, NodeState, Range};
```
gsserge, line 7: Missing newline.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+/// A closed interval range [min, max] used to describe a
+/// memory slot that will be assigned to a device by the VMM.
+/// This structure represents the key of the Node object in
+/// the interval tree implementation.
+pub struct Range {
+    pub min: u64,
+    pub max: u64,
+}
+
+// Node state for interval tree nodes.
```
gsserge, line 75:
```suggestion
/// Node state for interval tree nodes.
```

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,78 @@
...
+/// - Free -> Allocated: IntervalTree::allocate()
+/// - Allocated -> Free: IntervalTree::free()
+/// - * -> None: IntervalTree::delete()
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, Eq, Ord)]
+pub enum NodeState {
+    /// Node is free.
+    Free,
+    /// Node is allocated.
+    Allocated,
+}
```
studychao, line 61: Do we need to store value in the node to help user know which device/data/.. is in specified memory region?

studychao (commented):
Hi Alex, thanks for your work and here is my comments 😄

src/allocation_engine/interval_tree.rs
studychao: @jiangliu Hi gerry, What's your opinion on this?

gsserge: Do we really need to store references to devices in intervals? I'd really like to avoid that, because if we do, then the address allocator intersects with and basically becomes `IoManager` from `vm-device`. We would end up with two places storing the same information which is not great from technical and design perspectives (e.g. keeping them in sync). It it my understanding that `vm-allocator` is a utility crate, used by a VMM implementation in conjunction with `vm-device`'s `IoManager`.

gsserge (dismissed):

src/allocation_engine/interval_tree.rs
studychao: @gsserge . Agree.
I'm good with the `NodeState` Enum now:)

studychao (dismissed):

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,78 @@
+// Copyright (C) 2022 Alibaba Cloud. All rights reserved.
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2
+
+/// Policy for resource allocation.
+#[derive(Copy, Clone, Debug)]
```
andreeaflorescu, line 6: We should add `PartialEq` as well. You can check the guidelines for APIs here: https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits. We can leave out Display for now, and Hash, but the other ones are trivial to add as we can just use derive.

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,78 @@
...
+pub enum AllocPolicy {
+    /// Allocate from the first matched entry.
```
andreeaflorescu, line 8: nit: `from the first matched entry` sounds a bit off. Maybe what we want to say is "allocate the first matched"?

src/allocation_engine/interval_tree.rs
andreeaflorescu: Same for Constraint.

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Resource allocation policy.
+    pub policy: AllocPolicy,
+}
+
+/// A closed interval range [min, max] used to describe a
+/// memory slot that will be assigned to a device by the VMM.
+/// This structure represents the key of the Node object in
+/// the interval tree implementation.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Hash)]
+pub struct Range {
```
andreeaflorescu, line 43: Why aren't we using the range when defining the `Constraint`?

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Node is allocated.
+    Allocated,
+}
+
+/// Internal tree node to implement interval tree.
+#[derive(Debug, PartialEq)]
+struct InnerNode {
+    /// Interval handled by this node.
+    key: Range,
+    /// Optional contained data, None if the node is free.
```
andreeaflorescu, line 68: This comment seems to be outdated.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
+# Design
+
+## Allocation engine
+
+This implementation uses an interval tree that is specialised for allocation of
+memory-mapped I/O and port I/O address space. The fields of the strctures
+defined will have semantic meaning for this context (e.g node state to indicate
+if a node in the tree is assigned or not to a device).
+
+We offer three options for placing a memory slot in the managed address space:
```
andreeaflorescu, line 10: FirstMatch and LastMatch don't need a position argument. Would it make sense then to have `ExactMatch` define the range that needs to be allocated? This will enable us to have the same signature for allocation functions because the difference in parameter is captured in the allocationa_policy field.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+insert the range described by the constraint at the first available position
+starting from the end of the managed address space.
+2. `FirstMatch` -> When using this allocation policy the allocator will try to
+insert the range described by the constraint at the first available position
+starting from the begining of the managed address space.
+3. `ExactMatch` -> When using this allocation policy the allocator will try to
+insert the range at the exact position described by the constraint, otherwise
+it will return an error.
+
+```rust
```
andreeaflorescu, line 22: This kind of docs tend to get outdated really fast because we'll forget to update the docs when updating the code. If it doesn't bring much value, I would suggest we remove them from the document, especially since you are already describing the contents very well in the previous text block.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+    pub(crate) height: u64,
+    /// Cached maximum valued covered by this node.
+    pub(crate) max_key: u64,
+}
+```
+
+## Usage
+
+The concept of Interval Tree may seem complicated, but using vm-allocator to do
+resource allocation and release is simple and straightforward.
```
andreeaflorescu, line 109: This phrase does not contain any useful information. I would delete it.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+        address_allocator
+            .interval_tree
+            .insert(Range::new(base, end)?, NodeState::Free)?;
+        Ok(address_allocator)
+    }
+```
+
+After, the user should create a constraint with the size for the resource, optionally
+ the constraint could also contain the maximum, minimum and alignment for the
+ constraint.
```
andreeaflorescu, line 140: nit: added whitespace.

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+        };
+
+        address_allocator
+            .interval_tree
+            .insert(Range::new(base, end)?, NodeState::Free)?;
+        Ok(address_allocator)
+    }
+```
+
+After, the user should create a constraint with the size for the resource, optionally
```
andreeaflorescu, line 138: nit: we can break the phrase before "optionally".

src/allocation_engine/README.md
```diff
@@ -0,0 +1,170 @@
...
+        let constraint = Constraint::new(size).align(alignment).policy(alloc_policy);
+        // some other code here
+        if logical_condition {
+            let key = self.interval_tree.find_candidate(&constaint);
+            self.interval_tree.insert(key, NodeState::Allocated);
+        }
+    }
+```
+
+![Allocation example](/images/allocation_example.png)
```
andreeaflorescu, line 159: I don't really understand the idea of the image. What do the big boxes represent? What about the arrows? Maybe we can sync offline to clarify that a bit.

andreeaflorescu (changes requested):
Chatted with @AlexandruCihodaru offline. Before merging the PR:
- update the allocation_example.png to make it more explicit in terms of state transition and the operations that are generating the change
- include in the design document the design choices and the reasoning behind them: we want to use an interval tree for performance reasons, what are these reasons? where might it be used?
- remove the structure definitions from the design document and replace them with text where that makes sense. This helps with having mostly up to date documentation even when implementation details are changing.

For later (i.e. other PRs):
- examples are better to be placed in src/lib.rs because they are validated at runtime and they don't get so easily outdated.

src/allocation_engine/README.md
AlexandruCihodaru: I really like this suggestion.

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,76 @@
...
+        AllocPolicy::FirstMatch
+    }
+}
+
+/// Struct to describe resource allocation constraints.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct Constraint {
+    /// Size to allocate.
+    pub size: u64,
+    /// Range where the allocated resource will be placed.__rust_force_expr!
```
andreeaflorescu, line 28: nit: paste gone wrong with "rust_force_expr"?

andreeaflorescu (dismissed):
There are quite a few typos and other spell errors in the README. Can you please address them in a future PR?

gsserge (dismissed):

gsserge (approved):

src/allocation_engine/interval_tree.rs
```diff
@@ -0,0 +1,76 @@
...
+// SPDX-License-Identifier: Apache-2
+
+/// Policy for resource allocation.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum AllocPolicy {
+    /// Allocate the first matched entry.
+    FirstMatch,
+    /// Allocate first matched entry from the end of the range.
+    LastMatch,
+    /// Allocate a memory slot starting with the specified addrress
```
studychao, line 12: addrress -> address

gsserge (approved):

andreeaflorescu (approved):

PR #14, opened 2022-02-13
Added some of the methods provided by the defined structures.

AlexandruCihodaru:
* Added methods for the `Range` structure. The most notable ones are `align_forward` and `align_back` which are used during the slot allocation step.
* Added auxiliary methods for `NodeState` structure.


src/allocation_engine/interval_tree.rs
```diff
@@ -37,12 +40,93 @@ pub struct Constraint {
...
+        Ok(Range { min: min, max: max })
+    }
+
+    /// Get length of the range.
+    pub fn len(&self) -> u64 {
+        self.max - self.min + 1
+    }
+
+    /// Check whether two Range objects intersect with each other.
+    pub fn intersect(&self, other: &Range) -> bool {
```
andreeaflorescu, line 64: nit: I think you typically use `overlaps` as the name for this operation.

src/allocation_engine/interval_tree.rs
```diff
@@ -37,12 +40,93 @@ pub struct Constraint {
...
+        max(self.min, other.min) <= min(self.max, other.max)
+    }
+
+    /// Check whether the key is fully covered.
+    pub fn contain(&self, other: &Range) -> bool {
```
andreeaflorescu, line 69: These are typically named `contains`.

src/allocation_engine/interval_tree.rs
```diff
@@ -37,12 +40,93 @@ pub struct Constraint {
...
+        self.min <= other.min && self.max >= other.max
+    }
+
+    /// Create a new Range object with min being the first value from the range
```
andreeaflorescu, line 73: `min` is not very descriptive because it means knowing the implementation details of the structure `Range`. We can replace it with "the start of the region".

src/allocation_engine/interval_tree.rs
```diff
@@ -142,6 +142,16 @@ pub enum NodeState {
     Allocated,
 }
 
+impl NodeState {
+    fn replace(&mut self, value: NodeState) -> Self {
```
andreeaflorescu, line 146: How would this be used? Since it's an enum, why not just set another value directly?

src/allocation_engine/interval_tree.rs
```diff
@@ -37,12 +40,93 @@ pub struct Constraint {
...
     pub max: u64,
 }
 
+impl Range {
+    /// Create a new Range object.
+    pub fn new(min: u64, max: u64) -> Result<Self> {
+        if min > max {
+            return Err(Error::InvalidRange(min, max));
+        }
+        Ok(Range { min: min, max: max })
```
andreeaflorescu, line 55: This can be written as `Ok(Range { min, max })`.

src/allocation_engine/interval_tree.rs
```diff
@@ -37,12 +40,93 @@ pub struct Constraint {
...
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Hash, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Hash, Ord, Debug)]
 pub struct Range {
     pub min: u64,
     pub max: u64,
 }
 
+impl Range {
+    /// Create a new Range object.
+    pub fn new(min: u64, max: u64) -> Result<Self> {
```
andreeaflorescu, line 51: A `range` is an interval, so it's better defined with `start` and `end` instead of min/max. It also makes the code easier to read later on.

If we have a constructor, we should make `min` and `max` fields private (and rename them to start/end as well).

andreeaflorescu (commented):
Part 1. Still have to take a look at the align methods.

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,139 @@ pub struct Constraint {
...
+    pub fn align_forward(&self, align: u64) -> Result<Range> {
+        match align {
+            0 | 1 => Ok(*self),
+            _ => {
+                // Check alignment to be a power of two.
+                if !align.is_power_of_two() {
+                    return Err(Error::InvalidAlignment);
+                }
+                // !(align - 1)) will set all bits starting with the one already
+                // set in the alignment value, all bits until that one will be 0.
```
andreeaflorescu, line 106: I don't understand this phrase. Maybe instead of describing what `align - 1` does in text, we can just have a variable called `align_mask`. What would be important to not here is that this property only holds true for values that are a power of 2.

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,139 @@ pub struct Constraint {
...
 /// the interval tree implementation.
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Hash, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Hash, Ord, Debug)]
 pub struct Range {
-    pub min: u64,
-    pub max: u64,
+    /// Lower boundary of the interval.
+    range_start: u64,
+    /// Upper boundary of the interval.
+    range_end: u64,
```
andreeaflorescu, line 49: `range_` is superfluous. You are already in the `range` definition, so all the fields are already part of that namespace. `start` and `end` should suffice.

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,139 @@ pub struct Constraint {
...
+        if !align.is_power_of_two() {
+            return Err(Error::InvalidAlignment);
+        }
+
+        // The desired memory slot is represented by a range that starts at the
+        // highest multiple of align that is less than or equal to `candidate_address`.
+        // The highest multiple of n less than or equal to b is given by the
+        // equation floor(b/n) * n. If the result is less than the lower bound
+        // of the range, then no such multiple exists.
+        candidate_address = (candidate_address / align) * align;
```
andreeaflorescu, line 164: This operation is inefficient. There is no reason to use division, and we can compute the aligned address with bits operations similar to how it's done for the align_up.

andreeaflorescu (changes requested):
To have efficient operations when computing align_up and align_down, we need a helper function.

To compute the address aligned down, we need to:
- `addr & !(align - 1)`, for all values that are > 0 and a power of 2 -> let's call this `align_down`

To compute the align up, we need to:
- call `align_down` with `addr + align - 1`. This will insure that the operation is correct for already aligned values as well.

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,139 @@ pub struct Constraint {
...
+            return Err(Error::InvalidRange(range_start, range_end));
+        }
+        Ok(Range {
+            range_start,
+            range_end,
+        })
+    }
+
+    /// Get length of the range.
+    pub fn length(&self) -> u64 {
```
gsserge, line 65: `len()` is probably more idiomatic.

src/allocation_engine/interval_tree.rs
```diff
@@ -37,10 +41,125 @@ pub struct Constraint {
...
+    /// use vm_allocator::Result;
+    ///
+    /// fn intervals_align_forward() -> Result<Range> {
+    ///     let range_to_align = Range::new(2, 6)?;
+    ///     let res = range_to_align.align_forward(2)?;
+    ///     assert_eq!(res.length(), 5);
+    ///     Ok(res)
+    /// }
+    ///
+    /// intervals_align_forward().unwrap();
```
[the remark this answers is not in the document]
gsserge: I've read this doc comments a few times and I still don't understand what this method does.

I think it would be helpful to keep the example in the doc, but show only three lines (and not have the `intervals_align_forward` function at all) to clearly demonstrate how the aligned range looks like
```rust
let range = Range::new(2, 6).unwrap();
let aligned_range = range.align_forward(2).unwrap();
assert_eq!(aligned_range, Range::new(X, Y).unwrap());
```

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Clippy is gonna complain if we have defined len but not `is_empty` and in this context is_empty doesn't have sense in my opinion.

lauralt: @AlexandruCihodaru you can also ignore the clippy complaint with `#[allow(clippy::len_without_is_empty)]` above the `impl Range {`.

studychao (commented):
Hi Alex, following is my comments.

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,117 @@ pub struct Constraint {
...
+///         }
+///         return Err(Error::UnalignedAddress);
+///     }
+///     Err(Error::UnalignedAddress)
+/// }
+///
+/// # intervals_align_up().unwrap();
+/// ```
+pub fn align_up(address: u64, alignment: u64) -> Result<u64> {
+    if alignment == 0 {
```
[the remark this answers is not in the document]
AlexandruCihodaru: Yes this is correct, but we have to check that alignment is not 0 in this methods as well, because we call `align_down` with `address + (alignment - 1)`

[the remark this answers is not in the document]
studychao: Got that ~

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,117 @@ pub struct Constraint {
...
+        return Err(Error::InvalidAlignment);
+    }
+    if let Some(intermediary_address) = address.checked_add(alignment - 1) {
+        return align_down(intermediary_address, alignment);
+    }
+    Err(Error::UnalignedAddress)
```
studychao, line 150: The error name here may be not clear enough since the problem here is that the address is overflow instead of unaligned.

src/allocation_engine/interval_tree.rs
```diff
@@ -33,14 +37,117 @@ pub struct Constraint {
...
+        max(self.start, other.start) <= min(self.end, other.end)
+    }
+
+    /// Check whether the key is fully covered.
+    pub fn contains(&self, other: &Range) -> bool {
+        self.start <= other.start && self.end >= other.end
+    }
+
+    /// Get the beginning of the range.
+    pub fn get_start(&self) -> u64 {
```
andreeaflorescu, line 78: Rust getters don't use the keyword `get`. You should write directly `start` and `end`.

src/allocation_engine/README.md
```diff
@@ -67,6 +67,53 @@ they will be merged in a single node.
...
+0b0000000000000000000000000000000001000000000000000001000000000000(0x0000000040001000)
+```
+
+If the allocation policy used is `AllocPolicy::LastMatch` all memory slots will
+start at an address aligned to the first multiple of specified alignment value
+that is lower or equal to the candidate address.
+
+Example:
+
+```rust
```
andreeaflorescu, line 103: This does not seem right. The code block is marked as rust, but that would not compile.

studychao (approved):

andreeaflorescu (approved):

PR #16, opened 2022-02-22
Partial InnerNode implementation

AlexandruCihodaru:
### Summary of the PR

* Added search methods for the tree implementation.
* Added balancing methods for the tree implementation.

### Requirements

requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and follow
  [the 50/72 git commit rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule).
  test.

### DIFF
--- a/src/allocation_engine/interval_tree.rs
+++ b/src/allocation_engine/interval_tree.rs
@@ -229,6 +229,142 @@ impl InnerNode {
                 .and_then(|node| node.search_superset(key))
         }
     }
+
+    /// Rotates the tree such that height difference between subtrees
+    /// is not greater than abs(1).
+    #[allow(dead_code)]
+    fn rotate(self: Box<Self>) -> Box<Self> {
+        let l = height(&self.left);
+        let r = height(&self.right);
+
+        match (l as i64) - (r as i64) {
+            1 | 0 | -1 => self,
+            // Safe to unwrap as rotate_left_successor always returns Some when
+            // the current node has a left child and we just checked that it
+            // has at least one child otherwise this difference would not be two.
+            2 => self.rotate_left_successor().unwrap(),
+            // Safe to unwrap as rotate_right_successor always returns Some when
+            // the current node has a right child and we just checked that it
+            // has at least one child otherwise this difference would not be
+            // minus two.
+            -2 => self.rotate_right_successor().unwrap(),
+            _ => unreachable!(),
+        }
+    }
+
+    /// Performs a single left rotation on this node.
+    #[allow(dead_code)]
+    fn rotate_left(mut self: Box<Self>) -> Option<Box<Self>> {
+        if let Some(mut new_root) = self.right.take() {
+            self.right = new_root.left.take();
+            self.update_cached_height();
+            new_root.left = Some(self);
+            new_root.update_cached_height();
+            return Some(new_root);
+        }
+        None
+    }
+
+    /// Performs a single right rotation on this node.
+    #[allow(dead_code)]
+    fn rotate_right(mut self: Box<Self>) -> Option<Box<Self>> {
+        if let Some(mut new_root) = self.left.take() {
+            self.left = new_root.right.take();
+            self.update_cached_height();
+            new_root.right = Some(self);
+            new_root.update_cached_height();
+            return Some(new_root);
+        }
+        None
+    }
+
+    /// Performs a rotation when the left successor is too high.
+    #[allow(dead_code)]
+    fn rotate_left_successor(mut self: Box<Self>) -> Option<Box<Self>> {
+        if let Some(left) = self.left.take() {
+            if height(&left.left) < height(&left.right) {
+                self.left = left.rotate_left();
+                self.update_cached_height();
+            } else {
+                self.left = Some(left);
+            }
+            return self.rotate_right();
+        }
+        None
+    }
+
+    /// Performs a rotation when the right successor is too high.
+    #[allow(dead_code)]
+    fn rotate_right_successor(mut self: Box<Self>) -> Option<Box<Self>> {
+        if let Some(right) = self.right.take() {
+            if height(&right.left) > height(&right.right) {
+                self.right = right.rotate_right();
+                self.update_cached_height();
+            } else {
+                self.right = Some(right);
+            }
+            return self.rotate_left();
+        }
+        None
+    }
+
+    /// Deletes the entry point of this tree structure.
+    #[allow(dead_code)]
+    fn delete_root(mut self) -> Option<Box<Self>> {
+        match (self.left.take(), self.right.take()) {
+            (None, None) => None,
+            (Some(l), None) => Some(l),
+            (None, Some(r)) => Some(r),
+            (Some(l), Some(r)) => Some(Self::combine_subtrees(l, r)),
+        }
+    }
+
+    /// Finds the minimal key below the tree and returns a new optional tree
+    /// where the minimal value has been removed and the (optional) minimal node
+    /// as tuple (min_node, remaining).
+    #[allow(dead_code)]
+    fn get_new_root(mut self: Box<Self>) -> (Box<Self>, Option<Box<Self>>) {
+        match self.left.take() {
+            None => {
+                let remaining = self.right.take();
+                (self, remaining)
+            }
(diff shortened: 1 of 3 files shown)


src/allocation_engine/interval_tree.rs
```diff
@@ -185,8 +185,173 @@ struct InnerNode {
...
+
+    /// Returns a readonly reference to the node associated with the `key` or
+    /// None if there is no Node representing an interval that covers the
+    /// searched key. For a key [a, b], this method will return a node with
+    /// a key [c, d] such that c <= a and b <= d.
+    #[allow(dead_code)]
+    fn search_superset(&self, key: &Range) -> Option<&InnerNode> {
+        if self.key.contains(key) {
+            Some(self)
+        } else if key.end < self.key.start && self.left.is_some() {
```
gsserge, line 222: I wonder if we can use the same trick `.as_ref().and_then()` as in `search()` and get rid of unwraps and `else { None }`. Something like
```rust
if self.key.contains(key) {
    Some(self)
} else if key.end < self.key.start {
    self.left.as_ref().and_then(|node| node.search_superset(key))
} else { // seems like the only possibility here is key.start > self.key.end
    self.right.as_ref().and_then(|node| node.search_superset(key))
}
```

src/allocation_engine/interval_tree.rs
```diff
@@ -185,8 +185,173 @@ struct InnerNode {
...
+    }
+
+    /// Rotate the tree such that height difference between subtrees
+    /// is not greater than abs(1).
+    #[allow(dead_code)]
+    fn rotate(self) -> Box<Self> {
+        let l = height(&self.left);
+        let r = height(&self.right);
+
+        match (l as i32) - (r as i32) {
```
gsserge, line 240: Why cast to `i32`?

src/allocation_engine/interval_tree.rs
```diff
@@ -185,8 +185,173 @@ struct InnerNode {
...
+        let (mut new_root, remaining) = r.get_new_root();
+        new_root.left = Some(Box::new(l));
+        new_root.right = remaining;
+        new_root.updated_node()
+    }
+
+    /// Update cached information of the node.
+    /// Please make sure that the cached values of both children are up to date.
+    #[allow(dead_code)]
+    fn update_cached_info(&mut self) {
```
gsserge, line 339: If only the height is cached, then maybe it's better to be more specific with the method's name and call it `update_cached_height`. `info` sounds a bit generic.

src/allocation_engine/interval_tree.rs
```diff
@@ -185,8 +185,173 @@ struct InnerNode {
...
+        }
+    }
+
+    /// Perform a single left rotation on this node.
+    #[allow(dead_code)]
+    fn rotate_left(mut self) -> Box<Self> {
+        let mut new_root = self.right.take().expect("Node is broken");
+        self.right = new_root.left.take();
+        self.update_cached_info();
+        new_root.left = Some(Box::new(self));
```
gsserge, line 254: I'm a bit concerned with the amount of `Box::new()` calls in the `rotate*` methods. Each call is a memory allocation and is relatively expensive. Conceptually, you need to allocate memory only when instantiating a new node. The rotations should be performed by adjusting pointers, there should be no need for more allocations.

Now, I know that Rust is being, well, Rust, and simple things are usually difficult :) But could you please spend some time and investigate if it's possible to perform the rotations without allocations.

gsserge (dismissed):

studychao (commented):
Nice work:) So far so good, but I'll take a look at it again next morning.

src/allocation_engine/interval_tree.rs
```diff
@@ -286,4 +451,157 @@ mod tests {
...
+        right_node1.height = 4;
+        right_node2.height = 3;
+        right_node3.height = 2;
+
+        right_node3.right = Some(Box::new(right_node4));
+        right_node2.right = Some(Box::new(right_node3));
+        right_node1.right = Some(Box::new(right_node2));
+        balanced_tree.left = Some(Box::new(left_node1));
+        balanced_tree.right = Some(Box::new(right_node1));
+        balanced_tree = *balanced_tree.delete_root().unwrap();
```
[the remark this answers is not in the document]
AlexandruCihodaru: Exactly the unit tests are going to be changed when the insert method will be created

src/allocation_engine/interval_tree.rs
```diff
@@ -185,8 +185,173 @@ struct InnerNode {
...
+
+    /// Updates cached information of the node.
+    #[allow(dead_code)]
+    fn update_cached_height(&mut self) {
+        self.height = max(height(&self.left), height(&self.right)) + 1;
+    }
+
+    /// Updates the sub-tree to keep balance.
+    #[allow(dead_code)]
+    fn updated_node(mut self: Box<Self>) -> Box<Self> {
```
studychao, line 345: not sure the meaning of `updated` here, maybe this should be `update_node`

src/allocation_engine/interval_tree.rs
```diff
@@ -286,4 +451,158 @@ mod tests {
...
+        );
+        assert_eq!(
+            root_node.search_superset(&Range::new(0x1ff, 0x300).unwrap()),
+            None
+        );
+    }
+
+    #[test]
+    fn test_delete_root_balance_right() {
+        let mut unbalanced_tree =
```
studychao, line 515: I think `delete_root` is defined to be used on balanced_tree.
For example, if you add other larger node (e.g. `right_node5`) on the right subtree, the delete_root will panic because the height diffence during rotation is larger than 2.

src/allocation_engine/interval_tree.rs
```diff
@@ -286,4 +451,158 @@ mod tests {
...
+
+    #[test]
+    fn test_search_superset() {
+        let left_child = InnerNode::new(Range::new(0x90, 0x99).unwrap(), NodeState::Free);
+        let right_child = InnerNode::new(Range::new(0x200, 0x2ff).unwrap(), NodeState::Free);
+        let mut root_node = InnerNode::new(Range::new(0x100, 0x110).unwrap(), NodeState::Allocated);
+        root_node.left = Some(Box::new(left_child));
+        root_node.right = Some(Box::new(right_child.clone()));
+
+        assert_eq!(
```
andreeaflorescu, line 479: Can you write a bit about the test cases here? There are a lot of asserts and it's a bit hard to follow.

src/allocation_engine/interval_tree.rs
```diff
@@ -185,8 +185,173 @@ struct InnerNode {
...
+            2 => self.rotate_left_successor(),
+            -2 => self.rotate_right_successor(),
+            _ => unreachable!(),
+        }
+    }
+
+    /// Performs a single left rotation on this node.
+    #[allow(dead_code)]
+    fn rotate_left(mut self: Box<Self>) -> Box<Self> {
+        let mut new_root = self.right.take().expect("Node is broken");
```
andreeaflorescu, line 251: Why is this unwrap safe? Can you add a comment? We shouldn't have unwraps in the code unless they can be caused only by programming errors.

src/allocation_engine/interval_tree.rs
andreeaflorescu: s/updated_node/update_node

src/allocation_engine/interval_tree.rs
studychao: After discussion with Alex, `test_delete_root_*` are temporary unit tests for `delete_root` function and they will be replaced later when `insert` and `delete` for interval tree are introduced.

src/allocation_engine/interval_tree.rs
andreeaflorescu: Reading the comments, I think it's actually safer (in the sense that it's less prone to mistakes in the future) if we just return an `Option` here, and do the unwrap in the function where we validated that the child exists. The risk is that someone makes this function public, and then it can be called without doing the previously mentioned validation.

AlexandruCihodaru: Okay. Will update it

studychao (approved):

gsserge (approved):

PR #18, opened 2022-03-03
Tree insert, delete, update methods

AlexandruCihodaru:
### Summary of the PR

Addred methods for
* inserting a new node
* updating an existing node
* deleting a node
These methods will be used in a subsequent PR when memory allocation logic will be added.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/allocation_engine/interval_tree.rs
+++ b/src/allocation_engine/interval_tree.rs
@@ -332,7 +332,8 @@ impl InnerNode {
             Some(left) => {
                 let (min_node, left) = left.get_new_root();
                 self.left = left;
-                (min_node, Some(self.update_node()))
+                self.update_cached_height();
+                (min_node, Some(self.rotate()))
             }
         }
     }
@@ -344,7 +345,8 @@ impl InnerNode {
         let (mut new_root, remaining) = r.get_new_root();
         new_root.left = Some(l);
         new_root.right = remaining;
-        new_root.update_node()
+        new_root.update_cached_height();
+        new_root.rotate()
     }
 
     /// Updates cached information of the node.
@@ -353,11 +355,51 @@ impl InnerNode {
         self.height = max(height(&self.left), height(&self.right)) + 1;
     }
 
-    /// Updates the sub-tree to keep balance.
+    /// Insert a new node in the subtree. After the node is inserted the
+    /// tree will be balanced. The node_state parameter is needed because in
+    /// the AddressAllocator allocation logic we will need to insert both free
+    /// and allocated nodes.
     #[allow(dead_code)]
-    fn update_node(mut self: Box<Self>) -> Box<Self> {
+    pub(crate) fn insert(
+        mut self: Box<Self>,
+        key: Range,
+        node_state: NodeState,
+    ) -> Result<Box<Self>> {
+        // The InnerNode structure has 48 a length of 48 bytes. With other nested
+        // calls that are made during the insertion process the size occupied
+        // on the stack by just one insert call is around 122 bytes. Considering
+        // that the default stack size on Linux is 8K we could make around 73
+        // calls to insert method before confronting with an stack overflow. To
+        // be cautious we will use 50 as the maximum height of the tree. A
+        // maximum height of 50 will result in the possibility to allocate
+        // (2^50 - 1) memory slots. Considering the imposed maximum height the
+        // recursion is safe to use.
+        if (self.height + 1) > 50 {
+            return Err(Error::Overflow);
+        }
+        if self.key.overlaps(&key) {
+            return Err(Error::Overlap(key, self.key));
+        }
+        match self.key.cmp(&key) {
+            // It is not possible for a range to be equal with an existing node
+            // as the overlaps method will also catch this case and return the
+            // corresponding error code.
+            Ordering::Equal => unreachable!(),
+            Ordering::Less => match self.right {
+                None => self.right = Some(Box::new(InnerNode::new(key, node_state))),
+                Some(right) => {
+                    self.right = Some(right.insert(key, node_state)?);
+                }
+            },
+            Ordering::Greater => match self.left {
+                None => self.left = Some(Box::new(InnerNode::new(key, node_state))),
+                Some(left) => {
+                    self.left = Some(left.insert(key, node_state)?);
+                }
+            },
+        }
         self.update_cached_height();
-        self.rotate()
+        Ok(self.rotate())
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/allocation_engine/interval_tree.rs
```diff
@@ -359,6 +361,106 @@ impl InnerNode {
...
+            Ordering::Equal => {
+                return Err(Error::ResourceExhausted);
+            }
+            Ordering::Less => match self.right {
+                None => self.right = Some(Box::new(InnerNode::new(key, node_state))),
+                Some(_) => {
+                    self.right = self
+                        .right
+                        .take()
+                        .map(|n| n.insert(key, node_state).unwrap())
```
alsrdn, line 386: We should return an error here instead of unwrapping. There may be a case where a duplicate node is being inserted at a lower level in the subtree, which will cause a panic in this case instead of gracefully returning.

src/allocation_engine/interval_tree.rs
```diff
@@ -359,6 +361,106 @@ impl InnerNode {
...
+                }
+            },
+            Ordering::Greater => match self.left {
+                None => self.left = Some(Box::new(InnerNode::new(key, node_state))),
+                Some(_) => self.left = self.left.take().map(|n| n.insert(key, node_state).unwrap()),
```
alsrdn, line 391: Same comment as above about `unwrap()`

alsrdn (changes requested):

src/allocation_engine/interval_tree.rs
```diff
@@ -163,6 +163,8 @@ pub enum NodeState {
     Free,
     /// Node is allocated.
     Allocated,
+    /// Node does not exists
+    Inexistent,
```
gsserge, line 167: I'd say either `Nonexistent` or `DoesNotExist`.

src/allocation_engine/interval_tree.rs
```diff
@@ -359,6 +359,38 @@ impl InnerNode {
         self.update_cached_height();
         self.rotate()
     }
+
+    /// Insert a new node in the subtree. After the node is inserted the
+    /// tree will be balanced.
+    #[allow(dead_code)]
+    pub(crate) fn insert(
```
andreeaflorescu, line 368: How is this function going to be used?

There are 2 potential problems here:
1. Using recursion can generate a stack overflow, and it should be used with some defense mechanisms in place. One good option is to make sure you have a maximum possible depth. If the code can be written easily with iterations, than this should be the preferred method in production use cases.
2. Why is the `node_state` need to be passed for insert? Can we insert both allocated and free nodes?

src/allocation_engine/interval_tree.rs
```diff
@@ -467,142 +499,178 @@ mod tests {
...
             None
         );
         assert_eq!(
-            root_node.search_superset(&Range::new(0x1ff, 0x300).unwrap()),
+            tree.search_superset(&Range::new(0x1ff, 0x300).unwrap()),
             None
         );
     }
 
+    fn is_balanced(tree: Option<Box<InnerNode>>) -> bool {
```
andreeaflorescu, line 636: How is this going to be used? It would be better to have this as a function of `InnerNode` (i.e. receiving a reference to `&self` as a parameter). It's also a bit strange that the function receives an `Option` as a parameter because for `None` we shouldn't be asking the question about the tree being balanced or not. I guess this was added such that we can use recursion, but the problem can be solved with loops as well.

src/allocation_engine/interval_tree.rs
```diff
@@ -388,8 +390,39 @@ impl InnerNode {
...
             },
         }
-        Ok(self.update_node())
+        Ok(self.update_height_and_balance())
     }
+
+    /// Update the state of an old node. This method should be used when we
+    /// want to allocate a free node.
+    #[allow(dead_code)]
+    pub(crate) fn update(&mut self, key: &Range, node_state: NodeState) -> Result<()> {
```
andreeaflorescu, line 399: You are mentioning in the commit message that the only valid transition is from free to allocated. In this case, we can make this function impossible to be used wrong, and as a consequence simplify the code. This will eliminate some transitions, making the code less prone to error.

A few changes would be needed:
- remove the node_state parameter, we're only going to set the `node_state` to Allocated
- rename the function such that it suggests the transition is from free to allocated
- check that the node we want to update is Free

src/allocation_engine/interval_tree.rs
```diff
@@ -424,6 +424,34 @@ impl InnerNode {
...
+    pub(crate) fn delete(mut self: Box<Self>, key: &Range) -> Option<Box<Self>> {
+        match self.key.cmp(key) {
+            Ordering::Equal => {
+                return self.delete_root();
+            }
+            Ordering::Less => {
+                if let Some(node) = self.right.take() {
+                    let right = node.delete(key);
+                    self.right = right;
+                    return Some(self.update_height_and_balance());
```
andreeaflorescu, line 442: nit: functions named with `and` are typically a yellow flag because it means we're not following the single responsibility principle at the level of that function. In this case, it would be better to just call `update_height` and `rebalance` even if there are two calls.

andreeaflorescu (commented):
Scanned through the code, I'll still need to do a more in-depth review.

src/lib.rs
```diff
@@ -27,15 +27,27 @@ pub enum Error {
     /// An id  that was never allocated was requested to be released.
     #[error("Specified id: {0} was never allocated, can't release it.")]
     NeverAllocated(u32),
-    /// There are no more IDs available in the manage range
+    /// There resource we want to use or update is not available.
```
studychao, line 30: `There resource` should be a typo here

src/allocation_engine/interval_tree.rs
studychao: I have the same idea with this method :) Previously the `update` here will also update the data stored in the Node, but as we remove the design that every node stores some data, this part should also have some slight changes.

src/allocation_engine/interval_tree.rs
```diff
@@ -355,10 +357,101 @@ impl InnerNode {
...
+    pub(crate) fn insert(
+        mut self: Box<Self>,
+        key: Range,
+        node_state: NodeState,
+    ) -> Result<Box<Self>> {
+        if self.key.overlaps(&key) && self.key != key {
+            return Err(Error::Overlap(key, self.key));
+        }
+        match self.key.cmp(&key) {
+            Ordering::Equal => {
```
studychao, line 377: Is this  `Ordering::Equal` having the repeated judgment as to the above `self.key != key`?

src/allocation_engine/interval_tree.rs
```diff
@@ -355,10 +357,101 @@ impl InnerNode {
...
+                return Err(Error::ResourceNotAvailable);
```
studychao, line 378: In `self.key != key` situation, the error given is `Overlap`
and In `Ordering::Equal` situation, the error given is `ResourceNotAvailable`
I think we should find a unified way to describe the error when the range given is the same with some node's range.

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Indeed, we can also insert in tree a free node.

AlexandruCihodaru: I have addressed your comment regarding the recursion safety. The iterative code looks ugly and imho would be harder to maintain. I have computed how much a insert call should take on the stack and determined a max height. @raduiliescu had a really nice proposal to try to determine the stack size dynamically and to set the height on the run. While this is a nice proposal, I think we need an unsafe code block to do this.

studychao (approved):

src/allocation_engine/interval_tree.rs
andreeaflorescu: The `node_state` is still passed as a parameter. Can you remove it? Or otherwise explain why you need it?

src/allocation_engine/interval_tree.rs
```diff
@@ -373,7 +373,7 @@ impl InnerNode {
         // (2^50 - 1) memory slots. Considering the imposed maximum height the
         // recursion is safe to use.
         if (self.height + 1) > 50 {
-            return Err(Error::ResourceExhausted);
+            return Err(Error::ResourceNotAvailable);
```
andreeaflorescu, line 376: nit: I think `Overflow` might be a better error here because we cannot insert it because the tree height is too big.

src/allocation_engine/interval_tree.rs
```diff
@@ -399,6 +399,30 @@ impl InnerNode {
         self.update_cached_height();
         Ok(self.rotate())
     }
+
+    /// Update the state of an old node. This method should be used when we
```
andreeaflorescu, line 403: Hmmm I don't understand this part. The `insert` method already receives the `node_state` as a parameter. Why would you use this function then?

src/allocation_engine/interval_tree.rs
andreeaflorescu: This still has the `node_state` parameter. Can you either explain why it's needed or remove it?

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Changed it to Overflow.

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Updated the description to clarify things.

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Updated the comment to explain why we need to have the `node_state` parameter

src/allocation_engine/interval_tree.rs
```diff
@@ -353,11 +355,106 @@ impl InnerNode {
...
+        }
         self.update_cached_height();
-        self.rotate()
+        Ok(self.rotate())
+    }
+
+    /// Update the state of an old node. This method should be used when we
+    /// find an existing node with the state `NodeState::Free` that satisfies
+    /// all constraints of an allocation request. The recursion is safe as we
+    /// have in place a maximum height for the tree.
```
andreeaflorescu, line 408: We shouldn't have comments about how things are implemented in the doc comments. This should instead be written as a normal comment, using `//`.

andreeaflorescu (approved):

studychao (approved):

PR #19, opened 2022-04-06
Address allocator

AlexandruCihodaru:
### Summary of the PR

Added needed methods and structures for allocating memory slots using the AddressAllocator with the IntervalTree backend.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/address_allocator.rs
+++ b/src/address_allocator.rs
@@ -1,0 +1,243 @@
+// Copyright © 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright © 2022 Alibaba Cloud. All rights reserved.
+// Copyright © 2019 Intel Corporation. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Provides allocation and releasing strategy for memory slots.
+//!
+//! This module implements an allocation strategies for memory slots in an
+//! address space (for example MMIO and PIO).
+
+use crate::allocation_engine::IntervalTree;
+use crate::{AllocPolicy, Constraint, Error, Range, Result};
+
+// Internal representation of AddressAllocator. Contains the managed address
+// space represented through an instance of Range. The address allocator also
+// contains a node that represents the root of the interval tree used for
+// memory slots management. The reason we chose to use an interval tree is
+// that the average complexity for deletion and insertion is O(log N) and for
+// searching a node is O(N).
+/// Defines strategies for address space management. It exposes methods to
+/// allocate and free memory slots.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct AddressAllocator {
+    // Address space that we want to manage.
+    address_space: Range,
+    // Internal representation of the managed address space. Each node in the
+    // tree will represent a memory location and can have two states either
+    // `NodeState::Free` or `NodeState::Allocated`.
+    interval_tree: IntervalTree,
+}
+
+impl AddressAllocator {
+    /// Creates a new instance of AddressAllocator that will be used to manage
+    /// the allocation and release of memory slots from the managed address
+    /// space.
+    pub fn new(base: u64, size: u64) -> std::result::Result<Self, Error> {
+        let end = base
+            .checked_add(size.checked_sub(1).ok_or(Error::Underflow)?)
+            .ok_or(Error::Overflow)?;
+        let aux_range = Range::new(base, end)?;
+        Ok(AddressAllocator {
+            address_space: aux_range,
+            interval_tree: IntervalTree::new(aux_range),
+        })
+    }
+
+    /// Allocates a new aligned memory slot.
+    pub fn allocate(
+        &mut self,
+        size: u64,
+        alignment: u64,
+        alloc_policy: AllocPolicy,
+    ) -> Result<Range> {
+        if size == 0 {
+            return Err(Error::InvalidSize(size));
+        }
+
+        if !alignment.is_power_of_two() || alignment == 0 {
+            return Err(Error::InvalidAlignment);
+        }
+        let constraint = Constraint::new(size)?
+            .set_align(alignment)?
+            .set_policy(alloc_policy)?;
+        self.interval_tree.allocate(constraint)
+    }
+
+    /// Deletes the specified memory slot or returns `ResourceNotAvailable` if
+    /// the node was not allocated before.
+    pub fn free(&mut self, key: &Range) -> Result<()> {
+        self.interval_tree.free(key)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new_fails_overflow() {
+        assert_eq!(
+            AddressAllocator::new(u64::max_value(), 0x100).unwrap_err(),
+            Error::Overflow
+        );
+    }
+
+    #[test]
+    fn test_new_fails_size_zero() {
+        assert_eq!(
+            AddressAllocator::new(0x1000, 0x0).unwrap_err(),
+            Error::Underflow
+        );
+    }
+
+    #[test]
+    fn test_allocate_fails_alignment_zero() {
+        let mut pool = AddressAllocator::new(0x1000, 0x10000).unwrap();
+        assert_eq!(
+            pool.allocate(0x100, 0, AllocPolicy::FirstMatch)
+                .unwrap_err(),
+            Error::InvalidAlignment
+        );
+    }
+
+    #[test]
+    fn test_allocate_fails_alignment_non_power_of_two() {
+        let mut pool = AddressAllocator::new(0x1000, 0x10000).unwrap();
+        assert_eq!(
+            pool.allocate(0x100, 200, AllocPolicy::FirstMatch)
+                .unwrap_err(),
+            Error::InvalidAlignment
+        );
+    }
+
+    #[test]
+    fn test_allocate_fails_not_enough_space() {
+        let mut pool = AddressAllocator::new(0x1000, 0x1000).unwrap();
+        assert_eq!(
+            pool.allocate(0x800, 0x100, AllocPolicy::LastMatch).unwrap(),
+            Range::new(0x1800, 0x1FFF).unwrap()
+        );
+        assert_eq!(
+            pool.allocate(0x900, 0x100, AllocPolicy::FirstMatch)
+                .unwrap_err(),
+            Error::ResourceNotAvailable
+        );
+        assert_eq!(
+            pool.allocate(0x400, 0x100, AllocPolicy::FirstMatch)
+                .unwrap(),
+            Range::new(0x1000, 0x13FF).unwrap()
+        );
+    }
+
+    #[test]
+    fn test_allocate_with_alignment_first_ok() {
+        let mut pool = AddressAllocator::new(0x1000, 0x1000).unwrap();
+        assert_eq!(
+            pool.allocate(0x110, 0x100, AllocPolicy::FirstMatch)
+                .unwrap(),
+            Range::new(0x1000, 0x110F).unwrap()
+        );
+        assert_eq!(
+            pool.allocate(0x100, 0x100, AllocPolicy::FirstMatch)
+                .unwrap(),
+            Range::new(0x1200, 0x12FF).unwrap()
+        );
+        assert_eq!(
+            pool.allocate(0x10, 0x100, AllocPolicy::FirstMatch).unwrap(),
+            Range::new(0x1300, 0x130F).unwrap()
+        );
+    }
+
+    #[test]
+    fn test_allocate_with_alignment_last_ok() {
+        let mut pool_reverse = AddressAllocator::new(0x1000, 0x10000).unwrap();
+        assert_eq!(
+            pool_reverse
+                .allocate(0x110, 0x100, AllocPolicy::LastMatch)
+                .unwrap(),
+            Range::new(0x10E00, 0x10F0F).unwrap()
+        );
+        assert_eq!(
+            pool_reverse
+                .allocate(0x100, 0x100, AllocPolicy::LastMatch)
+                .unwrap(),
+            Range::new(0x10D00, 0x10DFF).unwrap()
+        );
+        assert_eq!(
+            pool_reverse
+                .allocate(0x10, 0x100, AllocPolicy::LastMatch)
+                .unwrap(),
+            Range::new(0x10C00, 0x10C0F).unwrap()
+        );
+    }
+
+    #[test]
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -39,12 +42,222 @@ pub enum Error {
...
+        max(self.start, other.start) <= min(self.end, other.end)
+    }
+
+    /// Check whether the key is fully covered.
+    pub fn contains(&self, other: &Range) -> bool {
+        self.start <= other.start && self.end >= other.end
+    }
+
+    /// Get the lower boundary of the range.
+    pub fn get_start(&self) -> u64 {
```
gsserge, line 95: style: With a few exceptions, the `get_` prefix is not used for getters in Rust code. [source](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)
Same for `get_end()`.

src/allocation_engine/interval_tree.rs
```diff
@@ -174,40 +96,57 @@ impl NodeState {
 
 /// Internal tree node to implement interval tree.
 #[derive(Clone, Debug, PartialEq, PartialOrd, Eq, Ord)]
-pub struct InnerNode {
+struct InnerNode {
     /// Interval handled by this node.
-    key: Range,
+    pub(crate) key: Range,
```
gsserge, line 101: I'm a bit concerned about such a high usage of `pub(crate)` in this module in combination with
```rust
/// Allocation engine used by address allocator.
pub mod allocation_engine;
```
in the crate root's `lib.rs`.

Of course this just my opinion, but in my mind the question of visibility is hierarchical and should not leak "throughout" the whole crate. Meaning that Interval Tree is an implementation detail and `interval_tree.rs` should be designed from the perspective of a standalone utility module, where the provided functionality is just `pub`.

The question of how this public functionality is used by the upper modules and whether it's exported to the end user should be decided by the upper level modules, `allocation_engine` and the top level `lib.rs`.

So, I'd rather make `interval_tree`'s functionality normally public, and control the end user experience in the top level `lib.rs` by exposing what's strictly needed - it seems like only `allocation_engine::NodeState` is needed because it's used in the `Error::InvalidStateTransition` variant.

gsserge:
Not related to the actual functionality, but I think it's useful to add `.gitignore` with at least `Cargo.lock` and `target/`.

gsserge (approved):

src/address_allocator.rs
```diff
@@ -0,0 +1,279 @@
...
+    // tree will represent a memory location and can have two states either
+    // `NodeState::Free` or `NodeState::Allocated`.
+    interval_tree: IntervalTree,
+}
+
+impl AddressAllocator {
+    /// Creates a new instance of AddressAllocator that will be used to manage
+    /// the allocation and release of memory slots from the managed address
+    /// space.
+    #[allow(dead_code)]
```
andreeaflorescu, line 34: Why do we still need to allow dead_code?

src/address_allocator.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn new(base: u64, size: u64) -> std::result::Result<Self, Error> {
+        if size == 0 {
+            return Err(Error::InvalidRange(base, base));
+        }
+        let end_option = base.checked_add(size - 1);
```
andreeaflorescu, line 39: nit: you can write this more rustacean by making use of `ok_or_else`:
```rust
        let end = base.checked_add(size - 1).ok_or_else(|| Error::InvalidRange(base, u64::MAX))?;
        Ok(AddressAllocator {
            address_space: Range::new(base, end)?,
            interval_tree: IntervalTree::new(aux_range, NodeState::Free),
        })
```

src/address_allocator.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+        Err(Error::InvalidRange(base, u64::MAX))
+    }
+
+    /// Allocates a new aligned memory slot.
+    #[allow(dead_code)]
+    pub fn allocate(
+        &mut self,
+        size: u64,
+        align_size: Option<u64>,
```
andreeaflorescu, line 55: nit: can you specify what is the default and why is this an option? Also, I think we should call this alignment instead of `align_size`.

src/lib.rs
```diff
@@ -39,12 +44,222 @@ pub enum Error {
...
+        Ok(Constraint {
+            size,
+            desired_range: Range::new(0, std::u64::MAX)?,
+            align: 4,
+            policy: AllocPolicy::default(),
+        })
+    }
+
+    /// Set the min constraint.
+    pub fn desired_range(mut self, range: Range) -> Self {
```
andreeaflorescu, line 142: There seems to be a mismatch between the name of the function and the doc comment. Which one is correct?

src/lib.rs
```diff
@@ -4,16 +4,21 @@
...
 
+mod address_allocator;
 /// Allocation engine used by address allocator.
-pub mod allocation_engine;
+mod allocation_engine;
 mod id_allocator;
 
+pub use crate::address_allocator::AddressAllocator;
+use crate::allocation_engine::NodeState;
 pub use crate::id_allocator::IdAllocator;
```
andreeaflorescu, line 14: In the IdAllocator doc comments there are some comments about the implementation. Those should be hidden. We can open a separate issue for that. I've spotted them by doing a `cargo doc --open`. Here are the comments that I am referring to: https://github.com/rust-vmm/vm-allocator/blob/ac13f051feef3c65ec76213770c87039cfd83871/src/id_allocator.rs#L14. These should be written with `//`, and we should have something more high level, concerning the user in the `///` type of comments.

src/lib.rs
```diff
@@ -39,12 +44,222 @@ pub enum Error {
...
+    }
+}
+
+/// Struct to describe resource allocation constraints.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct Constraint {
+    /// Size to allocate.
+    size: u64,
+    /// Range where the allocated resource will be placed.
+    desired_range: Range,
```
andreeaflorescu, line 123: nit: is this `exact_range`, as in a constraint through which we want to specify an exact allocation range instead of size or other constraints?

src/lib.rs
```diff
@@ -39,12 +44,222 @@ pub enum Error {
...
+    pub fn new(start: u64, end: u64) -> Result<Self> {
+        if start > end {
+            return Err(Error::InvalidRange(start, end));
+        }
+        Ok(Range { start, end })
+    }
+
+    /// Get length of the range.
+    pub fn len(&self) -> u64 {
+        self.end - self.start + 1
```
andreeaflorescu, line 83: The result of this computation is not correct when end and start are equal. It will output that the size is 1 instead of 0. Should we make this an invalid range in the constructor?

src/lib.rs
```diff
@@ -39,12 +44,222 @@ pub enum Error {
...
+    }
+
+    /// Get the upper boundary of the range.
+    pub fn end(&self) -> u64 {
+        self.end
+    }
+
+    /// Set the lower boundary of the range.
+    pub fn set_start(&mut self, start: u64) {
+        self.start = start;
```
andreeaflorescu, line 108: Should we check here that `start` is not higher than `end`? Otherwise we are allowing for wrong ranges to be created.

src/lib.rs
```diff
@@ -39,12 +44,222 @@ pub enum Error {
...
+    }
+
+    /// Set the upper boundary of the range.
+    pub fn set_end(&mut self, end: u64) {
```
andreeaflorescu, line 112: Should we also check here that end is not lower than start?

src/allocation_engine/interval_tree.rs
```diff
@@ -432,85 +332,364 @@ impl InnerNode {
...
-        assert!(range_a.contains(&Range::new(5u64, 6u64).unwrap()));
-        assert!(range_a.contains(&Range::new(6u64, 6u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(1u64, 1u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(1u64, 2u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(1u64, 3u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(1u64, 7u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(7u64, 8u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(6u64, 7u64).unwrap()));
-        assert!(!range_a.contains(&Range::new(7u64, 8u64).unwrap()));
+    #[allow(dead_code)]
```
andreeaflorescu, line 558: Why do we still need to allow dead code?

src/lib.rs
AlexandruCihodaru: I will create a new PR afterwards to update the documentation, comments, some tests and so on. Will create a separate issue and address these there. :D

src/lib.rs
```diff
@@ -39,12 +44,222 @@ pub enum Error {
...
+    start: u64,
+    /// Upper boundary of the interval.
+    end: u64,
+}
+
+#[allow(clippy::len_without_is_empty)]
+impl Range {
+    /// Create a new Range object.
+    pub fn new(start: u64, end: u64) -> Result<Self> {
+        if start > end {
```
gsserge, line 75: Related to the comment about `len()`: what if `start == end`? Should this also be an error?

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Because I have not added examples. Will create a new PR after this one to add examples, update documentation and READMEs and will remove the dead code then. Is this okay?

andreeaflorescu: The examples should not influence the `allow(dead_code)` macro. If there is still a warning from clippy regarding dead code, it means that we are not exporting things properly. Since from the implementation stand point this is complete, it would be better to remove `allow(dead_code)` now to make sure that the usage is not the one from the examples.

src/allocation_engine/interval_tree.rs
```diff
@@ -166,48 +44,66 @@ pub enum NodeState {
...
     /// Optional right child of current node.
-    right: Option<Box<InnerNode>>,
+    right: Option<Node>,
     /// Cached height of the node.
     height: u64,
 }
 
-impl InnerNode {
-    /// Creates a new InnerNode object.
+/// Wrapper structure over InnerNode, implements all methods needed to manage
```
andreeaflorescu, line 69: `InnerNode` is a private structure so we shouldn't mention it in public comments. We can simply say: "Wrapper over a node that implements the required functionality for managing an address space as interval trees".

src/allocation_engine/interval_tree.rs
```diff
@@ -166,48 +44,66 @@ pub enum NodeState {
...
+/// an address space with interval trees.
+#[derive(Clone, Debug, PartialEq, PartialOrd, Eq, Ord)]
+pub struct Node {
```
andreeaflorescu, line 72: What is the reason for having this separate structure?

Also, we can implement `Deref` for `Node` so we don't have to always write `self.inner_node.method_name`.

andreeaflorescu (commented):
I didn't get the time to take a look at the IntervalTree, but I'll be away until mid next week and we'll have no time to review the rest. Feel free to merge it without my formal approval.

gsserge (dismissed):

src/allocation_engine/interval_tree.rs
```diff
@@ -598,7 +699,7 @@ mod tests {
             .unwrap();
 
         assert_eq!(
-            tree.search_superset(&Range::new(0x100, 0x100).unwrap()),
+            tree.search_superset(&Range::new(0x100, 0x101).unwrap()),
```
alsrdn, line 702: Can you add some more documentation for these tests? I don't quite understand why you changed the values :)

src/lib.rs
alsrdn: After reading through the full implementation I understood that the `desired_range` is used to implement a limit on the addresses used for First and Last match.
Would it make more sense to move this range to the AllocPolicy?

alsrdn (changes requested):

src/allocation_engine/interval_tree.rs
AlexandruCihodaru: Because Range does not support having the start equal with the end anymore.

src/allocation_engine/interval_tree.rs
```diff
@@ -456,60 +334,308 @@ impl InnerNode {
...
+            }
+            AllocPolicy::ExactMatch(_) => unreachable!(),
+        };
+        // Create the result range.
+        let key = Range::new(
+            range_start,
+            std::cmp::min(self.key.end(), constraint.desired_range().end()),
+        )?;
+        // Check if the desired memory slot does fit in the candidate node.
+        if key.len() >= constraint.get_size() {
```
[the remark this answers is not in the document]
AlexandruCihodaru: We can not move this check exactly but we can add a new check 
```
if self.key.len() < constraint.get_size() {
    return some_error;
}
```

src/allocation_engine/interval_tree.rs
```diff
@@ -456,60 +334,308 @@ impl InnerNode {
...
+            // satisfying the specified constraints or `ResourceNotAvailable`
+            // if the request can not be satisfied.
+            AllocPolicy::LastMatch => self.last_match(constraint),
+            // Returns the node containing the address specified or the
+            // `ResourceNotAvailable` error if any of the sanity checks is not
+            // passing.
+            AllocPolicy::ExactMatch(start_address) => {
+                // If the desired start address is not aligned we do not
+                // continue and return UnalignedAddress error.
+                if start_address % constraint.get_align() != 0 {
```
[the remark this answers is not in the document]
AlexandruCihodaru: Don't see why we should double the AllocPolicy type check.

studychao (commented):
Hi Alex, here are my review comments. Please take a look at them.

Congrats for the final major PR for building vm-allocator:)

alsrdn (dismissed):

alsrdn (approved):

gsserge (approved):

studychao (approved):

src/lib.rs
AlexandruCihodaru: In the light of https://github.com/rust-vmm/vm-allocator/issues/42 it actually makes sense to have a cell of length of exactly 1.

src/lib.rs
AlexandruCihodaru: The result is actually what we were looking for, but I believe a more suitable name for this method would be `capacity`. WDYT?

andreeaflorescu: I think it's ok to leave it as `len`. This will also avoid breaking changes.

PR #21, opened 2022-04-19
[interval_tree]: ExactMatch change end address computation method to not miss a free slot.

AlexandruCihodaru:
When computing the end address in the `ExactMatch` we are looking for ranges greater than what was actually requested, this is caused by the fact that the ranges are inclusive on the left side so given a range [0x0 - 0x0FFF] that has the length 0x1000 and a allocation request with allocation policy `ExactMatch(0)` and size 0x1000 we will miss the said range because [here](https://github.com/rust-vmm/vm-allocator/blob/ce56e5c5db333d728547b6a990f916b46cdbb661/src/allocation_engine/interval_tree.rs#L342) the computed end address would be 0x1000 that does not belong to [0x0 - 0x0FFF] thus ignoring a valid range.

### Summary of the PR

* Updated the way we compute the end address in `AllocPolicy::ExactMatch` case, to reflect the fact that the Range is inclusive on the left side so the end address should be `start_address + size - 1`.
* Added regression test to check that we are not missing any available slot because of the off by one that was present before due to error described above
* Renamed `Range` structure to `RangeInclusive` to reflect that both ends of the range are inclusive as opposed to the std::ops::Range implementation.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/allocation_engine/interval_tree.rs
+++ b/src/allocation_engine/interval_tree.rs
@@ -340,14 +340,15 @@ impl InnerNode {
                     .search_superset(&Range::new(start_address, start_address + 1)?)
                     .ok_or(Error::ResourceNotAvailable)?;
                 let end_address = start_address
-                    .checked_add(constraint.size())
+                    .checked_add(constraint.size().checked_sub(1).ok_or(Error::Underflow)?)
                     .ok_or(Error::Overflow)?;
                 // We should check that starting from the desired address the
                 // whole memory slot will fit in the selected node.
-                if end_address > node.key.end() {
+                let range = Range::new(start_address, end_address)?;
+                if !node.key.contains(&range) {
                     return Err(Error::ResourceNotAvailable);
                 }
-                Ok((node, Range::new(start_address, end_address)?))
+                Ok((node, range))
             }
         }
     }
--- a/src/allocation_engine/interval_tree.rs
+++ b/src/allocation_engine/interval_tree.rs
@@ -2,7 +2,7 @@
 // Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2
 
-use crate::{AllocPolicy, Constraint, Error, Range, Result};
+use crate::{AllocPolicy, Constraint, Error, RangeInclusive, Result};
 use std::cmp::{max, Ordering};
 
 /// Returns the first multiple of `alignment` that is lower or equal to the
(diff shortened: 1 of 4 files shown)


src/allocation_engine/interval_tree.rs
```diff
@@ -344,7 +344,7 @@ impl InnerNode {
                     .ok_or(Error::Overflow)?;
                 // We should check that starting from the desired address the
                 // whole memory slot will fit in the selected node.
-                if end_address > node.key.end() {
+                if end_address.checked_sub(1).ok_or(Error::Underflow)? > node.key.end() {
```
andreeaflorescu, line 347: It looks like ranges in the crate are inconsistently defined:
- in this case, the range is non inclusive on the right end
- but in the `Range` definition, ranges are inclusive on both sides.

This is making the code hard to reason about, as you don't know out of the box with what kind of range you're dealing with.

andreeaflorescu: Chatted offline, resolving this. It looks like we're using Inclusive ranges everywhere. To set clear expectations about this, we're going to rename the structure from Range to RangeInclusive, thus also following the Rust std::ops::Range* defines.

gsserge:
I'm having a hard time understanding this PR :( The summary mentions some length check for `AllocPolicy::ExactMatch`, but it's not clear whether it's a check in the implementation or the tests. On top of that, figuring this out is complicated by the hundreds of lines of renaming of which nothing is said in the description.

AlexandruCihodaru:
> I'm having a hard time understanding this PR :( The summary mentions some length check for `AllocPolicy::ExactMatch`, but it's not clear whether it's a check in the implementation or the tests. On top of that, figuring this out is complicated by the hundreds of lines of renaming of which nothing is said in the description.

My bad as I have not updated the PR description, but I did now, please let me know if things are more clear now.

gsserge (dismissed):

src/allocation_engine/interval_tree.rs
```diff
@@ -340,11 +340,12 @@ impl InnerNode {
...
+                    .checked_add(constraint.size().checked_sub(1).ok_or(Error::Underflow)?)
                     .ok_or(Error::Overflow)?;
                 // We should check that starting from the desired address the
                 // whole memory slot will fit in the selected node.
-                if end_address > node.key.end() {
+                let range = Range::new(start_address, end_address)?;
+                if !node.key.contains(&range) {
                     return Err(Error::ResourceNotAvailable);
                 }
                 Ok((node, Range::new(start_address, end_address)?))
```
andreeaflorescu: The range is already created, we can just use it instead of recreating it here.

andreeaflorescu (dismissed):

andreeaflorescu (approved):

gsserge (approved):

PR #25, opened 2022-04-26
Replace additions and subtractions with `checked_add` and `checked_sub` methods

AlexandruCihodaru:
### Summary of the PR

Replaced additions and subtractions with `checked_add` and `checked_sub` methods for a safer computation of addresses.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/allocation_engine/interval_tree.rs
+++ b/src/allocation_engine/interval_tree.rs
@@ -12,6 +12,8 @@ pub fn align_down(address: u64, alignment: u64) -> Result<u64> {
     if !alignment.is_power_of_two() {
         return Err(Error::InvalidAlignment);
     }
+    // It is safe to subtract 1 as alignment is already checked to be greater
+    // than 0.
     Ok(address & !(alignment - 1))
 }
 
@@ -22,6 +24,8 @@ pub fn align_up(address: u64, alignment: u64) -> Result<u64> {
     if alignment == 0 {
         return Err(Error::InvalidAlignment);
     }
+    // It is safe to subtract 1 as alignment is already checked to be greater
+    // than 0.
     if let Some(intermediary_address) = address.checked_add(alignment - 1) {
         return align_down(intermediary_address, alignment);
     }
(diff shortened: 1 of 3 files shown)


src/allocation_engine/interval_tree.rs
```diff
@@ -12,6 +12,8 @@ pub fn align_down(address: u64, alignment: u64) -> Result<u64> {
     if !alignment.is_power_of_two() {
         return Err(Error::InvalidAlignment);
     }
+    // It is safe to subtract 1 as alignment is already checked to be greater
+    // than 0.
```
andreeaflorescu, line 16: I think this should be grater than 1. Same for the other comments.

src/lib.rs
```diff
@@ -78,7 +78,10 @@ pub struct RangeInclusive {
 impl RangeInclusive {
     /// Create a new RangeInclusive object.
     pub fn new(start: u64, end: u64) -> Result<Self> {
-        if start >= end {
+        // The length of the interval [0, u64::MAX] is u64::MAX + 1 which does
+        // not fit in a u64::MAX, hence we return `Error::InvalidRange` when
+        // there is an attempt to use that range.
+        if start >= end || (start == 0 && end == u64::MAX) {
```
alexandruag, line 84: This is not in the scope of the current PR, but was just wondering if we disallow singleton intervals (i.e. `[a,a]` by convention (hence the `start >= end` error convention instead of `start > end`), and if so do we always expect to need intervals with more than one element? Totally fine to address this in the future if there's something to address.

src/allocation_engine/interval_tree.rs
```diff
@@ -441,11 +452,18 @@ impl InnerNode {
         let range_start = match constraint.policy {
             AllocPolicy::FirstMatch => align_up(node_key.start(), constraint.align)?,
             AllocPolicy::LastMatch => {
+                // This operation can not underflow as we check at the beginning
```
alexandruag, line 455: Hmm, the second part of this paragraph is a bit hard to read ... is it basically saying that adding `1` towards the end of the chain cannot overflow because we previously subtracted a `size > 0` from the same value?

alexandruag (commented):
Looks good overall, just left a couple of questions.

alexandruag (approved):

PR #32, opened 2022-05-11
Update docs & fix bug in constraint

andreeaflorescu:
### Summary of the PR

This PR is the first in a series for publishing the crate on crates.io.
More details about the changes can be found in the commit messages.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/address_allocator.rs
+++ b/src/address_allocator.rs
@@ -17,8 +17,10 @@ use crate::{AllocPolicy, Constraint, Error, RangeInclusive, Result};
 // used for memory slots management. The reason we chose to use an interval tree
 // is that the average complexity for deletion and insertion is O(log N) and for
 // searching a node is O(N).
-/// Defines strategies for address space management. It exposes methods to
-/// allocate and free memory slots.
+/// An address space allocator.
+///
+/// The `AddressAllocator` manages an address space by exporting functionality to reserve and
+/// free address ranges based on a user defined [Allocation Policy](enum.AllocPolicy.html).
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub struct AddressAllocator {
     // Address space that we want to manage.
@@ -44,7 +46,7 @@ impl AddressAllocator {
         })
     }
 
-    /// Allocates a new aligned memory slot.
+    /// Allocates a new aligned memory slot. Returns the allocated range in case of success.
     pub fn allocate(
         &mut self,
         size: u64,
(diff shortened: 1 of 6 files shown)


gsserge (approved):

src/address_allocator.rs
```diff
@@ -47,22 +47,22 @@ impl AddressAllocator {
     }
 
     /// Allocates a new aligned memory slot. Returns the allocated range in case of success.
+    ///
+    /// When the `ExactMatch` policy is used, the start address MUST be aligned to the
+    /// alignment passed as a parameter.
+    ///
+    /// # Arguments:
+    /// - `size`: size to allocate.
+    /// - `align`: alignment to be used for the allocated resources.
```
lauralt, line 56: nit: the argument name is `alignment`.

src/address_allocator.rs
```diff
@@ -47,22 +47,22 @@ impl AddressAllocator {
...
+    ///            Valid alignments are a power of 2.
+    /// - `policy`: allocation policy.
```
lauralt, line 58: Same: policy -> alloc_policy.

src/lib.rs
```diff
@@ -17,6 +17,9 @@ use std::cmp::min;
 use std::result;
 use thiserror::Error;
 
+/// Default alignment that can be used for creating a `Constraint`.
+pub const DEFAULT_CONSTRAINT_ALIGN: u64 = 4;
```
lauralt, line 21: Do we also want to define a constructor that uses these defaults?

lauralt (approved):

src/lib.rs
andreeaflorescu: I am not sure what do you mean by that. Do you mean to have a new constructor where we don't have to pass the default alignment?

lauralt: The default align and the default policy; I noticed `Constraint::new(0x1000, 0x1000, AllocPolicy::default())` and I was thinking if it makes sense to add a `Default` impl for Constraint to shorten things a bit there. Not sure if it makes sense, I don't have much context on vm-allocator.

andreeaflorescu: I find that to be useful mostly in tests (i.e. creating a constraint with default values). I created a variable for the align policy mostly as an example, but we can also delete it. The code looked a bit different in my first iteration and I needed that.

lauralt (approved):

AlexandruCihodaru (approved):
LGTM 🚢

PR #36, opened 2022-05-17
newtype pattern

JonathanWoollett-Light:
### Summary of the PR

Implementing [the newtype pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html) for the [`vm_allocator::RangeInclusive`](https://docs.rs/vm-allocator/latest/vm_allocator/struct.RangeInclusive.html).

The newtype pattern is the typical approach to implementing different functionality on top of identical data, in this case we are implementing functionality on top of data identical to `std::ops::RangeInclusive`.

With the inability to now derive `PartialOrd` we write an explicit implantation that can be better understood without requiring prerequisite knowledge about how the derive macro for `PartialOrd` specifically works.

And implementing `Deref` allows easy access to the underlying `std::ops::RangeInclusive<u64>` removing the need to re-implement `start()` and `end()`. `Deref` could potentially introduce ambiguity between the function calls such as `std::ops::RangeInclusive::new()` and `vm_allocator::RangeInclusive::new()`, and `std::ops::RangeInclusive::contains()` and `vm_allocator::RangeInclusive::contains()` but it will always prefer the `vm_allocator::RangeInclusive` implementations (as can be seen in all tests continuing to pass) so this ambiguity would only have an affect upon readability. Although on this point I do not have a strong opinion and if you consider this ambiguity to be too much I could remove this.

Since now we cannot derive `Copy` it does add additional overhead of requiring explicit clones throughout the code, although this makes no functional difference.

I believe this change both brings usage more inline with recommended Rust patterns and increases the readability and explicitness of the code while introducing no meaningful functionality changes.

### Requirements

  description line.
  test.

### DIFF
--- a/src/address_allocator.rs
+++ b/src/address_allocator.rs
@@ -41,7 +41,7 @@ impl AddressAllocator {
             .ok_or(Error::Overflow)?;
         let aux_range = RangeInclusive::new(base, end)?;
         Ok(AddressAllocator {
-            address_space: aux_range,
+            address_space: aux_range.clone(),
             interval_tree: IntervalTree::new(aux_range),
         })
     }
--- a/src/allocation_engine/interval_tree.rs
+++ b/src/allocation_engine/interval_tree.rs
@@ -98,7 +98,7 @@ impl InnerNode {
     fn search_superset(&self, key: &RangeInclusive) -> Option<&InnerNode> {
         if self.key.contains(key) {
             Some(self)
-        } else if key.end < self.key.start {
+        } else if key.end() < self.key.start() {
             self.left
                 .as_ref()
                 .and_then(|node| node.search_superset(key))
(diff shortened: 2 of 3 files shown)


andreeaflorescu (commented):
I find the code to be superfluous with this change. I am not sure it was the best idea to suggest the use of `RangeInclusive`. The code probably would've looked different if RangeInclusive was used from the beginning, but right now readjusting it looks a bit of a hassle.

What do you think?

JonathanWoollett-Light:
> I find the code to be superfluous with this change

While it offers no obvious functional changes at this point since we are extending the functionality of an existing type the newtype pattern is the correct pattern to use in this instance improving the readability of the code (and better supports future changes such as using [`std::ops::Range`](https://doc.rust-lang.org/std/ops/struct.Range.html) underlying).

> I am not sure it was the best idea to suggest the use of `RangeInclusive`. The code probably would've looked different if RangeInclusive was used from the beginning, but right now readjusting it looks a bit of a hassle.

I would be in favor of using [`std::ops::Range`](https://doc.rust-lang.org/std/ops/struct.Range.html) instead but I believe this would come under a different PR whose changes can be made separately or following this one.

AlexandruCihodaru:
> I was thinking more of leaving the code as it is now instead of using RangeInclusive. If we want to switch to `Range`, than probably the code would become less bloated. Right now I don't see a strong advantage of using the new type pattern anymore. The code looks in this PR looks pretty complicated.

I endorse this.

PR #37, opened 2022-05-17
rustfmt.toml

JonathanWoollett-Light:
### Summary of the PR

Add `rustfmt.toml` to explicitly note the `rustfmt` configuration option `format_code_in_doc_comments=true` used.

The BuildKite CI runs `cargo fmt` with this option, it would be worth simply explicitly including in the repository for both ease of use and clarity.

### Requirements

  description line.
  test.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,1 @@
+format_code_in_doc_comments=true


andreeaflorescu:
@JonathanWoollett-Light can you add more details in the commit message. Since the CI already checks the format of the doc comments, it is not completely obvious why you are adding the change. You already explained this in the PR details, so it would be nice to also add this explanation in the commit message, as this is the one that can be checked when doing `git log`.

JonathanWoollett-Light:
I've add a short description of the changes in the commit message (if I should add more reasoning to it please say).

andreeaflorescu (approved):

PR #38, opened 2022-05-17
Use std::ops::RangeInclusive for IdAllocator

JonathanWoollett-Light:
### Summary of the PR

Instead of holding 2 `u32`s within `IdAllocator` to simulate a range, it could have a single value that is a range. This allows for easier construction, more explicit code and better readability.

This does introduce the need to dereference when getting the start and end of the range while somewhat awkward I don't believe this has any notable functional affect. This could be mitigated if `std::ops::Range` was used instead. I went with `std::ops::RangeInclusive` in this since it most closely resembles current usage and doesn't require many changes. I believe the change to `std::ops::Range` if it occurs should be in another commit to minimize too large changes in one commit/pr.

### Requirements

  description line.
  test.

### DIFF
--- a/src/id_allocator.rs
+++ b/src/id_allocator.rs
@@ -10,7 +10,8 @@
 use crate::{Error, Result};
 use std::collections::BTreeSet;
 
-/// An unique ID allocator that allows management of IDs in a given interval.
+/// An unique ID allocator that allows management of IDs in a given interval
+/// (inclusive).
 // Internal representation of IdAllocator. Contains the ends of the interval
 // that is managed, a field that points at the next available ID, and a
 // BTreeSet used to store the released IDs. The reason we chose a
@@ -19,12 +20,10 @@ use std::collections::BTreeSet;
 // are sorted so we will always use the first available ID.
 #[derive(Debug)]
 pub struct IdAllocator {
-    // Beginning of the range of IDs that we want to manage.
-    range_base: u32,
+    // The range of IDs that we want to manage.
+    range: std::ops::RangeInclusive<u32>,
     // First available id that was never allocated.
     next_id: Option<u32>,
-    // End of the range of IDs that we want to manage.
-    range_end: u32,
     // Set of all freed ids that can be reused at subsequent allocations.
     freed_ids: BTreeSet<u32>,
 }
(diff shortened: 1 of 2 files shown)


src/id_allocator.rs
```diff
@@ -19,35 +20,35 @@ use std::collections::BTreeSet;
...
-    /// `range_base` and `range_end`
-    pub fn new(range_base: u32, range_end: u32) -> Result<Self> {
-        if range_end < range_base {
-            return Err(Error::InvalidRange(range_base.into(), range_end.into()));
+    /// `range`.
+    pub fn new(range: std::ops::RangeInclusive<u32>) -> Result<Self> {
+        if range.end() < range.start() {
+            return Err(Error::InvalidRange(
+                *range.start() as u64,
+                *range.end() as u64,
```
andreeaflorescu, line 39: If ranges are based off u32, than the Invalid range should too I think. It is a bit strange to do the `as u64` conversion here. While this change was not introduced by your PR, it is made more obvious in it. Feel free to just open a separate issue for it if you don't want to address it in the same PR.

andreeaflorescu (commented):
Can you please add some more details about why you are doing this change in the commit summary? You can reuse the PR description.

src/id_allocator.rs
JonathanWoollett-Light: I noticed this but I was initially unsure if it was under the specific scope of this PR (as it would require changing the [`Error`](https://github.com/rust-vmm/vm-allocator/blob/beb21e6d1b32e03160446b7a326b80bbba6e7815/src/lib.rs#L93) type). If you think its in scope I'd be okay with making the changes here.

AlexandruCihodaru: From my PoV it is okay to change the Error enum in this PR as the modifications are needed by the present changes.

JonathanWoollett-Light: Having looked at surrounding code and reminding myself why I didn't change it, this is because [`RangeInclusive`](https://docs.rs/vm-allocator/latest/vm_allocator/struct.RangeInclusive.html) also produces an `Error::InvalidRange` under [`RangeInclusive::new`](https://docs.rs/vm-allocator/latest/vm_allocator/struct.RangeInclusive.html#method.new) this means error requires a possible range up to `u64`.

I could produce 2 `InvalidRange` variants one for `u32` and one for `u64` or I could use a generic, but either way in my opinion this becomes messier.

I think to do this well, requires some redesign and would warrant a separate PR.

PR #39, opened 2022-05-24
Unchecked implementations for IdAllocator

JonathanWoollett-Light:
### Summary of the PR

I have added unchecked implementations for `IdAllocator` ([`new_unchecked`](https://github.com/rust-vmm/vm-allocator/compare/main...JonathanWoollett-Light:id-allocator-unchecked?expand=1#diff-2b83e3b1a6a58e439c8b03c94bce0970d976ee55947654a11e4599666c1784aaR47-R57), [`allocate_id_unchecked`](https://github.com/rust-vmm/vm-allocator/compare/main...JonathanWoollett-Light:id-allocator-unchecked?expand=1#diff-2b83e3b1a6a58e439c8b03c94bce0970d976ee55947654a11e4599666c1784aaR101-R117) and [`free_id_unchecked`](https://github.com/rust-vmm/vm-allocator/compare/main...JonathanWoollett-Light:id-allocator-unchecked?expand=1#diff-2b83e3b1a6a58e439c8b03c94bce0970d976ee55947654a11e4599666c1784aaR142-R147) corresponding to the existing implementations `new`, `allocate_id` and `free_id` respectively). These unchecked implementations allow skipping certain checks offering better performance.

It is common to encounter circumstance where we can can guarantee that a given `Result` will always be `Ok` and thus a given `unwrap()` will never panic (often when using constants). In this circumstances it can be beneficial to avoid unnecessary checks, a couple examples from the standard library are [`i32::unchecked_add`](https://doc.rust-lang.org/std/primitive.i32.html#method.unchecked_add) and [`slice::get_unchecked`](https://doc.rust-lang.org/std/primitive.slice.html#method.get_unchecked).

In implementing these unchecked functions which mirror existing functions I have also made minor stylistic changes to the existing functions such that they better match each other. I have also mirrored all tests for the unchecked variants.

If it where already present in the project I would have added benchmarks to evaluate the performance of the checked vs unchecked variants, although I felt including a large dependency such as [criterion](https://crates.io/crates/criterion) (even as a `dev-dependency`) would warrant its own pr and should introduce benchmarks for most functionality in the library rather than a small subset.

---

The specific use case that brought up this suggestion was when using constants for a given range e.g.
```rust
pub const IRQ_MAX: u32 = 128;
/// First usable interrupt on aarch64.
pub const IRQ_BASE: u32 = 32;
```
When calling `IdAllocator::new(IRQ_BASE,IRQ_MAX).unwrap()` we are doing unnecessary checks within `IdAllocator::new`.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,10 @@ edition = "2018"
 [dependencies]
 libc = ">=0.2.39"
 thiserror = "1.0"
+
+[dev-dependencies]
+criterion = "0.3.5"
+
+[[bench]]
+name = "benchmarks"
+harness = false
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -1,0 +1,57 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use vm_allocator::*;
+
+const MIN: u32 = 0;
+const MAX: u32 = 100;
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function("IdAllocator::new", |b| {
+        b.iter(|| IdAllocator::new(MIN, MAX).unwrap())
+    });
+    c.bench_function("IdAllocator::new_unchecked", |b| {
+        b.iter(|| IdAllocator::new_unchecked(MIN, MAX))
+    });
+    c.bench_function("IdAllocator::allocate_id", move |b| {
+        b.iter_with_setup(
+            || IdAllocator::new(MIN, MAX).unwrap(),
+            |mut allocator| {
+                allocator.allocate_id().unwrap();
+            },
+        )
+    });
+    c.bench_function("IdAllocator::allocate_id_unchecked", |b| {
+        b.iter_with_setup(
+            || IdAllocator::new_unchecked(MIN, MAX),
+            |mut allocator| {
+                allocator.allocate_id_unchecked();
+            },
+        )
+    });
+    c.bench_function("IdAllocator::free_id", |b| {
+        b.iter_with_setup(
+            || {
+                let mut allocator = IdAllocator::new(MIN, MAX).unwrap();
+                allocator.allocate_id().unwrap();
+                allocator
+            },
+            |mut allocator| {
+                allocator.free_id(MIN).unwrap();
+            },
+        )
+    });
+    c.bench_function("IdAllocator::free_id_unchecked", |b| {
+        b.iter_with_setup(
+            || {
+                let mut allocator = IdAllocator::new_unchecked(MIN, MAX);
+                allocator.allocate_id_unchecked();
+                allocator
+            },
+            |mut allocator| {
+                allocator.free_id_unchecked(MIN);
+            },
+        )
+    });
+}
+
+criterion_group!(benches, criterion_benchmark);
+criterion_main!(benches);
--- a/src/id_allocator.rs
+++ b/src/id_allocator.rs
@@ -32,7 +32,7 @@ pub struct IdAllocator {
 impl IdAllocator {
     /// Creates a new instance of IdAllocator that will be used to manage the
     /// allocation and release of ids from the interval specified by
-    /// `range_base` and `range_end`
+    /// `range_base` and `range_end`.
     pub fn new(range_base: u32, range_end: u32) -> Result<Self> {
         if range_end < range_base {
             return Err(Error::InvalidRange(range_base.into(), range_end.into()));
@@ -44,6 +44,17 @@ impl IdAllocator {
             freed_ids: BTreeSet::<u32>::new(),
         })
     }
+    /// Creates a new instance of IdAllocator that will be used to manage the
+    /// allocation and release of ids from the interval specified by
+    /// `range_base` and `range_end`, without checking `range_end < range_base`.
+    pub fn new_unchecked(range_base: u32, range_end: u32) -> Self {
+        IdAllocator {
+            range_base,
+            next_id: Some(range_base),
+            range_end,
+            freed_ids: BTreeSet::<u32>::new(),
+        }
+    }
 
     fn id_in_range(&self, id: u32) -> bool {
         // Check for out of range.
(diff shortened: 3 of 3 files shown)


AlexandruCihodaru:
Hello @JonathanWoollett-Light thank you for your contribution. Can you present any data regarding the gained performance?
We have to check that the difference in performance justifies having unsafe API for the library.

**Example**: 
Considering that a VMM takes user input for VMs configuration, a malicious actor could request N devices (where N is greater than the upper boundary of the interval). In this case we gain some performance by not checking the generated ID but we lose it back when the virtualization backend fail to register the irq. I think that performance wise is better to fail early while generating the ID than to fail in the hypervisor after a context switch.

JonathanWoollett-Light:
### Results

Function|Average|Unchecked average|%|% speedup
---|---|---|---|---
`new`|13.104 ns|8.6105 ns|65.71|34.29
`allocate_id`|21.802 ns|20.542 ns|94.22|5.78
`free_id`|47.874 ns|45.810 ns|95.69|4.31

*Running on a c5.metal ec2 instance.*

A consistent speedup of >4% in frequently called functions is not entirely insignificant (even in the nanoseconds). 

Similarly [`i32::unchecked_add`](https://doc.rust-lang.org/std/primitive.i32.html#method.unchecked_add) and [`slice::get_unchecked`](https://doc.rust-lang.org/std/primitive.slice.html#method.get_unchecked) illustrate relatively minuscule performance gains yet still have good justification for existing as functions.

### Discussion

> Considering that a VMM takes user input for VMs configuration, a malicious actor could request N devices (where N is greater than the upper boundary of the interval). In this case we gain some performance by not checking the generated ID but we lose it back when the virtualization backend fail to register the irq. I think that performance wise is better to fail early while generating the ID than to fail in the hypervisor after a context switch.
 
I would always lean towards the users choice (in this case the VMM developer who depends on `vm-allocator`) as to whether they believe a function cannot fail. We only lose this performance if a failure would occur, I would argue regardless you should trust users by offering them the choice.

Much like `i32::unchecked_add` and `slice::get_unchecked` in the standard library, a developer can still choose to use `i32::checked_add` or `slice::get`.

But overall, in short, it is better to call `IdAllocator::new_unchecked(0, 100);` than `IdAllocator::new().unwrap(0, 100);` as they illustrate identical safety but one is slightly more performant (one can also imagine a circumstance in which this applies to `allocate_id` and `free_id`). It is the better (if only a very slightly better) and for this I would argue it is worth it.

---

### Full output

```
IdAllocator::new        time:   [13.090 ns 13.104 ns 13.121 ns]                              
                        change: [-0.2449% +0.0225% +0.3469%] (p = 0.90 > 0.05)
                        No change in performance detected.
Found 20 outliers among 100 measurements (20.00%)
  2 (2.00%) high mild
  18 (18.00%) high severe

IdAllocator::new_unchecked                                                                             
                        time:   [8.5974 ns 8.6105 ns 8.6266 ns]
                        change: [-0.3939% -0.1219% +0.1704%] (p = 0.42 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe

IdAllocator::allocate_id                                                                            
                        time:   [21.765 ns 21.802 ns 21.847 ns]
                        change: [+0.3905% +0.6708% +0.9802%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 15 outliers among 100 measurements (15.00%)
  12 (12.00%) high mild
  3 (3.00%) high severe

IdAllocator::allocate_id_unchecked                                                                            
                        time:   [20.530 ns 20.542 ns 20.561 ns]
                        change: [-0.2952% -0.2225% -0.1176%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 15 outliers among 100 measurements (15.00%)
  5 (5.00%) high mild
  10 (10.00%) high severe

IdAllocator::free_id    time:   [47.822 ns 47.874 ns 47.935 ns]                                 
                        change: [-3.2941% -3.0315% -2.7591%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe

IdAllocator::free_id_unchecked                                                                            
                        time:   [45.750 ns 45.810 ns 45.891 ns]
                        change: [-0.4611% -0.0780% +0.2977%] (p = 0.71 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
```

AlexandruCihodaru:
> ### Results
> Function	Average	Unchecked average	%	% speedup
> `new`	13.104 ns	8.6105 ns	65.71	34.29
> `allocate_id`	21.802 ns	20.542 ns	94.22	5.78
> `free_id`	47.874 ns	45.810 ns	95.69	4.31
> _Running on a c5.metal ec2 instance._
> 
> A consistent speedup of >4% in frequently called functions is not entirely insignificant (even in the nanoseconds).
> 
> Similarly [`i32::unchecked_add`](https://doc.rust-lang.org/std/primitive.i32.html#method.unchecked_add) and [`slice::get_unchecked`](https://doc.rust-lang.org/std/primitive.slice.html#method.get_unchecked) illustrate relatively minuscule performance gains yet still have good justification for existing as functions.
> 
> ### Discussion
> > Considering that a VMM takes user input for VMs configuration, a malicious actor could request N devices (where N is greater than the upper boundary of the interval). In this case we gain some performance by not checking the generated ID but we lose it back when the virtualization backend fail to register the irq. I think that performance wise is better to fail early while generating the ID than to fail in the hypervisor after a context switch.
> 
> I would always lean towards the users choice (in this case the VMM developer who depends on `vm-allocator`) as to whether they believe a function cannot fail. We only lose this performance if a failure would occur, I would argue regardless you should trust users by offering them the choice.
> 
> Much like `i32::unchecked_add` and `slice::get_unchecked` in the standard library, a developer can still choose to use `i32::checked_add` or `slice::get`.
> 
> But overall, in short, it is better to call `IdAllocator::new_unchecked(0, 100);` than `IdAllocator::new().unwrap(0, 100);` as they illustrate identical safety but one is slightly more performant (one can also imagine a circumstance in which this applies to `allocate_id` and `free_id`). It is the better (if only a very slightly better) and for this I would argue it is worth it.
> 
> ### Full output
> ```
> IdAllocator::new        time:   [13.090 ns 13.104 ns 13.121 ns]                              
>                         change: [-0.2449% +0.0225% +0.3469%] (p = 0.90 > 0.05)
>                         No change in performance detected.
> Found 20 outliers among 100 measurements (20.00%)
>   2 (2.00%) high mild
>   18 (18.00%) high severe
> 
> IdAllocator::new_unchecked                                                                             
>                         time:   [8.5974 ns 8.6105 ns 8.6266 ns]
>                         change: [-0.3939% -0.1219% +0.1704%] (p = 0.42 > 0.05)
>                         No change in performance detected.
> Found 5 outliers among 100 measurements (5.00%)
>   1 (1.00%) low mild
>   1 (1.00%) high mild
>   3 (3.00%) high severe
> 
> IdAllocator::allocate_id                                                                            
>                         time:   [21.765 ns 21.802 ns 21.847 ns]
>                         change: [+0.3905% +0.6708% +0.9802%] (p = 0.00 < 0.05)
>                         Change within noise threshold.
> Found 15 outliers among 100 measurements (15.00%)
>   12 (12.00%) high mild
>   3 (3.00%) high severe
> 
> IdAllocator::allocate_id_unchecked                                                                            
>                         time:   [20.530 ns 20.542 ns 20.561 ns]
>                         change: [-0.2952% -0.2225% -0.1176%] (p = 0.00 < 0.05)
>                         Change within noise threshold.
> Found 15 outliers among 100 measurements (15.00%)
>   5 (5.00%) high mild
>   10 (10.00%) high severe
> 
> IdAllocator::free_id    time:   [47.822 ns 47.874 ns 47.935 ns]                                 
>                         change: [-3.2941% -3.0315% -2.7591%] (p = 0.00 < 0.05)
>                         Performance has improved.
> Found 8 outliers among 100 measurements (8.00%)
>   4 (4.00%) high mild
>   4 (4.00%) high severe
> 
> IdAllocator::free_id_unchecked                                                                            
>                         time:   [45.750 ns 45.810 ns 45.891 ns]
>                         change: [-0.4611% -0.0780% +0.2977%] (p = 0.71 > 0.05)
>                         No change in performance detected.
> Found 10 outliers among 100 measurements (10.00%)
>   4 (4.00%) high mild
>   6 (6.00%) high severe
> ```
Considering that `allocate_id` method is called less than 100 times in a normal use case, that would bring a performance improvement of 0.00012222 seconds. Same for `free_id`. Regarding `new` method it is usually called one time per VM lifetime. From my personal point of view, maintenance effort is bigger than the performance gained. But please ping other maintainers as well to check their opinion.

andreeaflorescu:
I would prefer for these API updates to not be merged. My reasoning is the following:
- the existing API is not called often enough to make this performance improvements significant
- from the Firecracker's point of view, these API calls are also not on the critical performance path
- with the added API checks we would need to re-run the security related tests and check if any assumptions are invalidated

The PR though adds some other changes as well that could be nice to merge. I am talking about the refactoring of `allocate_id` using matches.

alsrdn:
@JonathanWoollett-Light I also would prefer not to merge these changes.
Apart from the concerns of @AlexandruCihodaru and @andreeaflorescu which I find totally valid, my concern is that using the crate with this new API diminishes it's functionality and purpose and leaves the crate user unprotected from more complicated errors.

I can give you a plausible scenario:
We have an IdAllocator that we use to allocate interrupt lines from.
The interrupt lines managed by the allocator are between 0 and 32. Interrupt lines 32-64 are used for some platform components (e.g. timers).
We have a user configuration that wants to create 33 devices. Each device interrupt line is allocated using `allocate_id_unchecked`. Because we made an error when checking the number of devices, we allocate interrupt line 33 and overwrite the assignment for line 33 to a different device than the intended one (e.g. a timer).
Now we get a silent error where a device will not work (the timer), instead of the software rejecting the configuration.

I think we need to keep a minimum set of assumptions for the functionality and keep APIs safe and robust.
The implementation of the IdAllocator is pretty simple; if someone wants to get the ultimate performance, a u32 increment can suffice (and is actually used in Firecracker today).

PR #56, opened 2023-04-05
Enable dependabot for Cargo.toml

roypat:
### Summary of the PR

Enables dependabot for `Cargo.toml`. Note that since this is a library target, it will only open pull requests for semver incompatible updates (e.g. those that `cargo update` would not do).

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,8 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+- package-ecosystem: cargo
+  directory: "/"
+  schedule: 
+    interval: weekly
+  open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -4,4 +4,9 @@ updates:
...
   schedule:
     interval: weekly
-  open-pull-requests-limit: 10
\ No newline at end of file
+  open-pull-requests-limit: 10
+- package-ecosystem: cargo
+  directory: "/"
+  schedule: 
+    interval: weekly
+  open-pull-requests-limit: 10
```
lauralt, line 12: I think we should also add [this](https://github.com/rust-vmm/vhost-device/blob/main/.github/dependabot.yml#L8). More info here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow.

lauralt: Can you also leave a link for this part: `it will only open pull requests for semver incompatible updates`. if that's the case I was wondering if it will ever open PRs for indirect deps. What I would like to happen is that dependabot will open PRs (or at least announce us) if an indirect dep is vulnerable even if it is a compatible update.
This might be of interest in this type of scenario: https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates and https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#managing-dependabot-security-updates-for-your-repositories.

andreeaflorescu, line 12: We might want to limit this to 1 instead because otherwise dependabot will open a lot of PRs at the same time, and then because of conflicts we will not be able to merge them anyway. Limiting this at 1 would allow us to progressively merge all the dependency updates without conflicts.

roypat: Regarding indirect dependencies: Since vm-allocator is a library crate, I don't think that applies. There is no Cargo.lock, so indirect dependencies are not tracked anywhere. Its a good thing to keep in mind for binary crates though, so thanks for pointing that option out to me!
For the second part, that was mostly me testing in my fork (e.g. adding new entries to Cargo.toml and seeing what dependabot ends up creating PRs for. It did not open any for outdated patch versions, or for outdated minor versions if the major is >= 1)

roypat: Dependabot will automatically resolve conflicts whenever a PR is merged (and if that fails then you can always manually tell it to regenerate a PR). Since there's only a Cargo.toml here, all dependabot PRs should be `-1/+1` PRs, on disjoint lines, so there shouldn't be merge problems. 

I'd prefer to have dependabot open all the PRs at once to get an overview (especially because I'm not sure what delay there will be between merging one PR and dependabot opening the next one). Is dependabot already enabled on any other rust-vmm crate? We should strive for consistency here, too.

andreeaflorescu: I didn't know that dependabot now is able to fix conflicts, interesting. The problem remains though that you will need to re-approve on push force, so even if dependabot fixes the conflict, you still need to manually go to re-approve them one by one. AFAIK, the only other repository where dependabot is enabled is the vhost-devices one, but there approvals are not needed to merge the PRs, which is something that they did with the purpose of not having to re-approve on push forces. This is not something that I would be comfortable doing for other critical rust-vmm components though. While I also like consistency, in this context I don't think we can apply it.

We can leave it as is and reconsider in the future if you feel strongly about it. Let's see how this works.

roypat: Yeah, I'd be in favor of observing how this plays out first. I'm hoping that dependabot is clever enough to realize that if a PR merges cleanly into main, it will not do the whole rebase-and-force-push, and instead recognize that the PR is mergable after clicking "update branch", which preserves approvals.

andreeaflorescu (approved):

.github/dependabot.yml
vireshk: IIRC, after one request is merged, if I let dependabot rebase the changes by itself (i.e. without clicking "update branch"), then it used to discard all the previous approvals. While clicking on "update branch" would create a merge commit and preserve the approvals.

I have gotten over it now, after I accidentally enabled "merge without approvals" somehow. I am using it only for bot requests which pass the tests. I simply go and rebase/merge the changes without waiting for anyone else's approvals, which are anyway not that important for bot requests I feel.

roypat: Aw, that's unfortunate :(. Thanks for letting me know! 

I suppose there would be the option to disable the automatic rebasing altogether and rely on "update branch", and if something can't be "update branch"'d then manually trigger a rebase with `@dependabot rebase` (because in those cases, we cant get around manual intervention anyway)

lauralt (approved):

.github/dependabot.yml
andreeaflorescu: @roypat can you update it to not do push forces in this case?

roypat: Alright, did this in a follow-up: https://github.com/rust-vmm/vm-allocator/pull/61

PR #120, opened 2026-05-29
id-allocator: fix double-alloc and add `is_allocated` public fn

JamesC1305:
- Fix an off-by-one error in free_id that could allow for ids to be double allocated
- Add new `is_allocated` method, so that implementing programs can query whether an id is currently allocated or not
- Update changelog accordingly

### Requirements

requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
  test.

### DIFF
--- a/src/id_allocator.rs
+++ b/src/id_allocator.rs
@@ -84,7 +84,7 @@ impl IdAllocator {
             return Err(Error::OutOfRange(id));
         }
         if let Some(next_id) = self.next_id {
-            if next_id < id {
+            if next_id <= id {
                 return Err(Error::NeverAllocated(id));
             }
         }
@@ -170,6 +170,21 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_free_id_never_allocated_boundary() {
+        let mut allocator = IdAllocator::new(5, 23).unwrap();
+
+        assert_eq!(allocator.next_id.unwrap(), 5);
+        assert_eq!(allocator.free_id(5).unwrap_err(), Error::NeverAllocated(5));
+        assert_eq!(allocator.freed_ids.len(), 0);
+
+        for _ in 0..3 {
+            allocator.allocate_id().unwrap();
+        }
+        assert_eq!(allocator.next_id.unwrap(), 8);
+        assert_eq!(allocator.free_id(8).unwrap_err(), Error::NeverAllocated(8));
+        assert_eq!(allocator.freed_ids.len(), 0);
+    }
     #[test]
     fn test_id_sanity_checks() {
         let legacy_irq_allocator = IdAllocator::new(5, 23).unwrap();
--- a/src/id_allocator.rs
+++ b/src/id_allocator.rs
@@ -76,6 +76,25 @@ impl IdAllocator {
         Err(Error::Overflow)
     }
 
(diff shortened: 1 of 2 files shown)


src/id_allocator.rs
```diff
@@ -192,6 +211,55 @@ mod tests {
...
+        assert_eq!(id, 5);
+        assert!(allocator.is_allocated(5));
+    }
+
+    #[test]
+    fn test_is_allocated_exhausted_range() {
+        // When the range is exhausted, `next_id` becomes `None`. Every id in
+        // the range has been handed out, so all of them are allocated until
+        // freed.
+        let mut allocator = IdAllocator::new(u32::MAX - 1, u32::MAX).unwrap();
```
ShadowCurse: Maybe use a bit smaller numbers like `IdAllocator::new(0, 1)`?

JamesC1305: Can do, but that covers a slightly different code path. The current implementation exercises the `None` branch of `is_allocated`'s match on `next_id`, which is only reached when `next_id` would overflow. Changing to `(0,1)` would instead exercise the `Some` path, but when `id < next == false`. 

Perhaps it's worth testing both paths? Same conclusion, but slightly different ways of getting there.

JamesC1305: Have separated to have tests for each cases, as well as a test for freeing an id in the middle of the range

ShadowCurse: ? the `next_id` will be None when the range is exhausted and for [0..1] range it will be after 2 allocations (just like in current test).
Anyway, this is just a nit, so feel free to ignore.

JamesC1305: That's what I thought, but that's not the case. If you have range [0..1], and allocate both 0 and 1, `next_id` will remain as `Some(2)`, but on the next allocation, `next_id` is compared aganist `self.range_end` and returns `Error::ResourceNotAvailable`. Only when `next_id.checked_add(1)` overflows is `next_id` set to `None`, and `Error::Overflow` is returned instead.

Let me know if I misinterpreted what you said

ShadowCurse: oh, sorry, for some reason I thought `next_id` would be `None` once the whole range is used, but it is only `None` for overflow case. In that case I guess you can do:
```rust
        assert_eq!(allocator.allocate_id().unwrap(), u32::MAX - 1);
        assert!(allocator.is_allocated(u32::MAX - 1)); <- tests `Some` case

        assert_eq!(allocator.allocate_id().unwrap(), u32::MAX);
        assert!(allocator.is_allocated(u32::MAX)); <- tests `None` case

        ...
```

JamesC1305: Okay, I changed it to match that, ~~but it starts from `u32::MAX - 2` instead~~. ~~The `Some` case has two subcases (`id < next_id` -> `true` returned and `id >= next_id` -> `false` returned), so we test both now + the overflow in the same test.`~~

Edit: -2 was not needed. We can test both sides of the sum case with -1

ShadowCurse (approved):

src/id_allocator.rs
ShadowCurse: nice, thx

andreeaflorescu (approved):

PR #121, opened 2026-06-03
chore: prepare v0.1.4 release

JamesC1305:
Prepare a new release, so the new `is_allocated` method from PR #120 is made available.

### Requirements

requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,19 +3,23 @@
 ## Upcoming version
 
 ### Added
+### Changed
+### Fixed
+### Removed
+### Deprecated
+
+## [v0.1.4]
+
+### Added
 
 - Support for no_std environments.
 - [[#120]](https://github.com/rust-vmm/vm-allocator/pull/120): `IdAllocator::is_allocated` to query whether a given id is currently
   allocated.
 
-### Changed
 ### Fixed
 
 - [[#120]](https://github.com/rust-vmm/vm-allocator/pull/120): `IdAllocator::free_id` incorrectly accepted freeing `next_id` (an id that was
   never allocated), which could lead to the same id being allocated twice.
-
-### Removed
-### Deprecated
 
 ## [v0.1.3]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "vm-allocator"
-version = "0.1.3"
+version = "0.1.4"
 description = "Helpers for allocating resources needed during the lifetime of a VM."
 repository = "https://github.com/rust-vmm/vm-allocator"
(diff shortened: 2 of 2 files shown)


ShadowCurse (dismissed):

CHANGELOG.md
```diff
@@ -2,18 +2,20 @@
...
-  allocated.
 
 ### Changed
 ### Fixed
-
-- [[#120]](https://github.com/rust-vmm/vm-allocator/pull/120): `IdAllocator::free_id` incorrectly accepted freeing `next_id` (an id that was
-  never allocated), which could lead to the same id being allocated twice.
-
 ### Removed
 ### Deprecated
```
andreeaflorescu: If a section is empty, can you remove it instead?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "vm-allocator"
-version = "0.1.3"
+version = "0.1.4"
```
andreeaflorescu, line 3: Is this really a patch release? I was thinking since it's adding support for no-std, it should be a minor instead.

JamesC1305: It's up to the maintainers I suppose, happy to bump the minor version instead if you think it warrants that. @ShadowCurse, any thoughts?

ShadowCurse: I think we should keep bumping of minor version for breaking changes. (since the crate is still not 1.0)

JamesC1305: @andreeaflorescu if that sounds okay to you, I can just update the changelog (in response to your other comment).

andreeaflorescu: Yes, please do.

JamesC1305: Done, ty!

ShadowCurse (approved):

Cargo.toml
andreeaflorescu: The version is still 0.1.4, which means we bumped the patch version, not the minor one. Is this what you meant @ShadowCurse? I thought we will move to 0.2.0.

andreeaflorescu (approved):
For added functionality we've been using minor versions even when dealing with 0.x.y releases in other crates. I see though that this is not the case with vm-allocator, so feel free to merge it as is.

Orthogonal question: what needs to happen for this to be 1.0?

Cargo.toml
ShadowCurse: Yes, minor for breaking changes, patch for additions.

ShadowCurse:
> what needs to happen for this to be 1.0?

Good question. I guess we can mark it 1.0 at any time we want. But for now I see no real reason to do it.

